PR #24, opened 2025-07-21
feat: add loader component cy-12

watchlar2000:
Hello there! Below is the list of the changes made:

- Created loader component that can be configured via props to control size, theme, and container (either used as a fullscreen feature or inside buttons).

```tsx
<Loader />   
<button>
  Click me <Loader container="inline" size="small" /> 
</button>                           
<Loader theme="accent" size="large" /> 
```

- Added loader styles into assets/css.
- Did a minor refactor inside css folder to create utilities folder - for now it's just visually-hidden.css inside it.

Below is a demo of how this loader component can be utilized. By default, it's a spinner in the center of the screen with grayish background (fullscreen). Background can be turned off by passing `backdrop={false}` prop. Loader also can be customized with size and theme (I didn't find any particular UI requirements for loader). To use it with buttons - `container="inline"` prop should be passed down.

Peek 2025-07-23 11-05

### DIFF
--- a/apps/frontend/src/assets/css/components/loader.css
+++ b/apps/frontend/src/assets/css/components/loader.css
@@ -1,0 +1,68 @@
+.loader {
+	--loader-base-size: 16px;
+}
+
+.loader-fullscreen {
+	z-index: 9999;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	height: 100vh;
+	background-color: #e0e0e09f;
+}
+
+.loader-inline {
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.loader-small {
+	--loader-base-size: 12px;
+}
+
+.loader-medium {
+	--loader-base-size: 20px;
+}
+
+.loader-large {
+	--loader-base-size: 32px;
+}
+
+.loader-brand {
+	color: #75a5be;
+}
+
+.loader-muted {
+	color: #dcf3b6;
+}
+
+.loader-accent {
+	color: #ff7171;
+}
+
+.loader-spinner {
+	display: inline-block;
+	width: var(--loader-base-size);
+	height: var(--loader-base-size);
+	border: 3px solid currentcolor;
+	border-right-color: transparent;
+	border-radius: 50%;
+	animation: spin 0.5s linear infinite;
+}
+
+@media (prefers-reduced-motion: reduce) {
+	.loader-spinner {
+		animation: spin 1.5s linear infinite;
+	}
+}
+
+@keyframes spin {
+	0% {
+		transform: rotate(0deg);
+	}
+
+	100% {
+		transform: rotate(360deg);
+	}
+}
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -63,9 +63,10 @@ const App: React.FC = () => {
 					width: "max-content",
 				}}
 			>
-				<Loader container="inline" />
+				<Loader container="inline" size="base" />
 				<span>Button with loader</span>
 			</button>
+
 			<div>
 				<RouterOutlet />
 			</div>
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -48,7 +48,24 @@ const App: React.FC = () => {
 				</li>
 			</ul>
 			<p>Current path: {pathname}</p>
-
+			<button
+				style={{
+					alignItems: "center",
+					backgroundColor: "#C6E7F9",
+					border: "2px solid #33322E",
+					borderRadius: "999px",
+					cursor: "pointer",
+					display: "flex",
+					fontSize: "18px",
+					fontWeight: "normal",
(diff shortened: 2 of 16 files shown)


apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -25,6 +25,10 @@ const App: React.FC = () => {
 		}
 	}, [isRoot, dispatch]);
 
+	if (dataStatus === DataStatus.PENDING) {
```
daniil-tkalenko, line 28: add isLoading variable, would be better for future changes and more readable
also check for DataStatus.IDLE status

daniil-tkalenko (changes requested):

apps/frontend/src/assets/css/components/loader.css
```diff
@@ -0,0 +1,68 @@
+.loader {
+	--loader-base-size: 16px;
+}
+
+.loader-fullscreen {
+	z-index: 9999;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	height: 100vh;
```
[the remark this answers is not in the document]
watchlar2000: Sure, that's would be a better way to organize the styles! Thanks!

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,30 @@
...
+const Loader = ({
+	container = "fullscreen",
+	size = "large",
+	theme = "brand",
+}: Properties): JSX.Element => {
+	return (
+		<div className={`loader loader-${container}`}>
+			<div
+				className={`loader-spinner loader-${theme} loader-${size}`}
+				role="status"
```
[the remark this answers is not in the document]
watchlar2000: Fixed this one as well. I've added clsx and created a helper - cn, a small abstraction over clsx. Will it make sense?

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -41,6 +48,24 @@ const App: React.FC = () => {
...
+					fontSize: "18px",
+					fontWeight: "normal",
+					gap: "12px",
+					padding: "8px 16px",
+					width: "max-content",
+				}}
+			>
+				<Loader container="inline" size="base" />
+				<span>Button with loader</span>
+			</button>
```
daniil-tkalenko, line 68: now sure what it is, if it's just for testing purpuse, you should remove that from PR after testing.
btw we will have common Button component, and it would be good to add loader to it is related property is present on the Button

daniil-tkalenko (changes requested):

WilhelmWesser:
You should add a screenshot of the loader component you created to the PR description so that reviewers know how it looks in UI

package.json
```diff
@@ -64,5 +64,8 @@
 	"simple-git-hooks": {
 		"pre-commit": "npm run lint:type && npm run git:hook:precommit",
 		"commit-msg": "npm run git:hook:commit"
+	},
+	"dependencies": {
+		"clsx": "2.1.1"
```
daniil-tkalenko, line 69: I think you should add it to frontend app instead of root

apps/frontend/src/libs/components/app/app.tsx
WilhelmWesser: I think it will be a scope of the task with adding a common button component

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,41 @@
...
+type LoaderTheme = "accent" | "brand" | "muted";
+
+type Properties = {
+	backdrop?: boolean;
+	container?: LoaderContainer;
+	size?: LoaderSize;
+	theme?: LoaderTheme;
+};
+
+const Loader = ({
```
WilhelmWesser, line 18: Let's add a type for the component as FC<Properties>

apps/frontend/src/libs/helpers/cn.ts
```diff
@@ -0,0 +1,7 @@
+import clsx, { type ClassValue } from "clsx";
+
+const cn = (...classes: ClassValue[]): string => {
```
WilhelmWesser, line 3: When I looked at the name of this function, I did not get what it does. Let's name it as `getValidClassNames`

package.json
WilhelmWesser, line 69: This library must be a part of the frontend workspace we do not need it for the whole project. I mean the package.json inside the frontend. Uninstall it and install again as `npm install clsx -w frontend`, and it will appear in the frontend workspace

WilhelmWesser (changes requested):

apps/frontend/src/libs/components/app/app.tsx
watchlar2000: That button is for demo purposes only - to showcase how button with loader will look like. Better remove it?

watchlar2000:
> You should add a screenshot of the loader component you created to the PR description so that reviewers know how it looks in UI

Added a pic with some usecases to the PR.

roman-mykytka (approved):

PR #25, opened 2025-07-22
feat(backend): implement fetch authenticated user functionality cy-11

alex-tuz:
Note: Currently accepts userId as URL parameter for testing purposes. After JWT middleware implementation, this will be changed to extract userId from JWT token in Authorization header, and endpoint will become GET /auth/profile (without :id parameter).


packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,4 +1,5 @@
 const AuthApiPath = {
+	PROFILE: "/profile/:id",
```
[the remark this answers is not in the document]
WilhelmWesser: Indeed, profile is not the best name here. May be let's name it as `authenticated-user`?
@alex-tuz @daniil-tkalenko what do you think?

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +13,21 @@ class AuthService {
...
 	}
 
+	public async getAuthenticatedUser(
+		userId: number,
+	): Promise<UserProfileResponseDto> {
+		const user = await this.userService.findById(userId);
+
+		if (!user) {
+			throw new HTTPError({
+				message: "User not found",
```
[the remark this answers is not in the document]
alex-tuz: Thanks for the suggestion. I’ve updated the code — now all messages use the ErrorMessage enum

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -81,6 +83,52 @@ class AuthController extends BaseController {
...
+			throw new HTTPError({
+				message: ErrorMessage.UNAUTHORIZED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		return {
+			payload: await this.authService.getAuthenticatedUser(user.id),
+			status: HTTPCode.OK,
+		};
```
daniil-tkalenko, line 131: basically you can return user from options and that's it
even error handling won't be needed since if it's not public route and there is no user - auth plugin from https://github.com/BinaryStudioAcademy/bsa-2025-checkly/pull/29 will throw this error

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -8,6 +10,12 @@ import {
 	type ControllerRouteParameters,
 } from "./libs/types/types.js";
 
+type CustomFastifyRequest = Parameters<
+	ServerApplicationRouteParameters["handler"]
+>[0] & {
```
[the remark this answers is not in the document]
daniil-tkalenko: @alex-tuz you most likly wont need this type once auth plugin merged btw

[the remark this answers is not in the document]
alex-tuz: > @alex-tuz you most likly wont need this type once auth plugin merged btw
For now, I marked my changes as TODO. Once the plugin is merged into the main branch, I'll remove the CustomFastifyRequest type definition - otherwise, I'm currently getting a type definition error.

apps/backend/src/modules/auth/auth.controller.ts
daniil-tkalenko: @alex-tuz not really, plugin in this case is a middleware, it should run each time somebody trying to access not public rote. it checks token presence in request and if there is no token - auth error will be thrown. So, if there are some issues with a token, we would not reach handler logic at all. I think we can relay on plugin in this case, without a need to add some back up code.

apps/backend/src/libs/modules/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,9 +1,12 @@
+import { type UserGetAllItemResponseDto } from "~/libs/types/types.js";
+
 type APIHandlerOptions<
 	T extends DefaultApiHandlerOptions = DefaultApiHandlerOptions,
 > = {
 	body: T["body"];
 	params: T["params"];
 	query: T["query"];
+	user?: undefined | UserGetAllItemResponseDto;
```
WilhelmWesser, line 9: If you have user? undefined is redudnant, remove `undefined`

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -81,6 +81,41 @@ class AuthController extends BaseController {
 				body: userSignUpValidationSchema,
 			},
 		});
+
+		this.addRoute({
+			handler: (options) => this.getAuthenticatedUser(options),
+			method: "GET",
```
WilhelmWesser, line 87: Use enum with request methods

WilhelmWesser (approved):

WilhelmWesser (approved):

roman-mykytka (approved):

apps/backend/src/libs/modules/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,9 +1,12 @@
+import { type UserGetAllItemResponseDto } from "~/libs/types/types.js";
+
 type APIHandlerOptions<
 	T extends DefaultApiHandlerOptions = DefaultApiHandlerOptions,
 > = {
 	body: T["body"];
 	params: T["params"];
 	query: T["query"];
+	user?: UserGetAllItemResponseDto;
```
daniil-tkalenko, line 9: lets use UserDto here

apps/backend/src/modules/users/user.service.ts
```diff
@@ -53,6 +54,12 @@ class UserService implements Service {
 		return await this.userRepository.findByField("email", email);
 	}
 
+	public async findById(id: number): Promise<null | UserGetAllItemResponseDto> {
```
daniil-tkalenko, line 57: and here

packages/shared/src/modules/users/libs/types/user-profile-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type UserProfileResponseDto = {
+	email: string;
+	id: number;
+	name: string;
+};
+
+export { type UserProfileResponseDto };
```
daniil-tkalenko, line 7: we can propaply remove this type, since it's not used right now, and duplicates UserDto

packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,4 +1,5 @@
 const AuthApiPath = {
+	ME: "/me",
```
daniil-tkalenko, line 2: please, also update frontend action where we fetch authed user, so it use your ME path instead of  PROFILE, and expects UserDto | null as a result

daniil-tkalenko (approved):

PR #26, opened 2025-07-22
feat: add protected route cy-8

olenkabilonizhka:
Implemented ProtectedRoute component.

- Redirect not signed in users from protected screens to Sign In page.

### DIFF
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -1,8 +1,31 @@
 import { ToastContainer } from "react-toastify";
 
 import { RouterOutlet } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/enums.js";
+import {
+	useAppDispatch,
+	useAppSelector,
+	useEffect,
+	useLocation,
+} from "~/libs/hooks/hooks.js";
+import { actions as userActions } from "~/modules/users/users.js";
 
 const App: React.FC = () => {
+	const { pathname } = useLocation();
+	const dispatch = useAppDispatch();
+	useAppSelector(({ users }) => ({
+		dataStatus: users.dataStatus,
+		users: users.users,
+	}));
+
+	const isRoot = pathname === AppRoute.ROOT;
+
+	useEffect(() => {
+		if (isRoot) {
+			void dispatch(userActions.loadAll());
+		}
+	}, [isRoot, dispatch]);
+
 	return (
 		<>
 			<RouterOutlet />


apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
+import { StorageKey } from "~/libs/modules/storage/storage.js";
+
+const ProtectedRoute: React.FC = () => {
+	const isAuthorized = !!localStorage.getItem(StorageKey.TOKEN);
+
+	return isAuthorized ? <Outlet /> : <Navigate replace to={AppRoute.SIGN_IN} />;
+};
+
+export { ProtectedRoute };
```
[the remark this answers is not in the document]
olenkabilonizhka: Thanks for the comments, I resolved it. Also I added a new type UserGetAuthenticatedResponseDto, which will be enhanced in issue 10.

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { DataStatus } from "~/libs/enums/enums.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+const ProtectedRoute: React.FC = () => {
+	const { dataStatus, user } = useAppSelector(({ auth }) => ({
+		dataStatus: auth.dataStatus,
+		user: auth.user,
+	}));
+
+	return dataStatus === DataStatus.FULFILLED && user ? (
```
daniil-tkalenko, line 13: would be better to create a variable for that check with a proper name

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const ProtectedRoute: React.FC = () => {
+	const { dataStatus, user } = useAppSelector(({ auth }) => ({
+		dataStatus: auth.dataStatus,
+		user: auth.user,
+	}));
+
+	const isAuthorized = dataStatus === DataStatus.FULFILLED && user;
+
+	return isAuthorized ? <Outlet /> : <Navigate replace to={AppRoute.SIGN_IN} />;
```
daniil-tkalenko, line 15: we probaply would need to check pending status as well and show Loader component in this case
if so I think it would be better to rewrite it and use early returns instead of ternary operator
otherwise LGTM

daniil-tkalenko (approved):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
WilhelmWesser: The Loader component has just been merged, so you can use it

olenkabilonizhka: I resolved it. Could you please review and merge my PR?

roman-mykytka (changes requested):

WilhelmWesser (approved):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	const isAuthorized = dataStatus === DataStatus.FULFILLED && user;
+
+	if (isAuthorized) {
+		return <Outlet />;
+	}
+
+	return <Navigate replace to={AppRoute.SIGN_IN} />;
+};
```
daniil-tkalenko, line 28: there was added hook to fetch authed user
lets move it to the ProtectedRoute from App component

apps/frontend/src/index.tsx
```diff
@@ -32,6 +33,16 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
...
 							},
+							{
+								children: [
+									{
+										element: "Dashboard",
+										path: AppRoute.DASHBOARD,
+									},
+								],
+								element: <ProtectedRoute />,
+								path: AppRoute.ROOT,
```
daniil-tkalenko, line 44: propaply we should use AppRoute.DASHBOARD here as well?

daniil-tkalenko (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
daniil-tkalenko: I would also suggest to remove Users and Status from the App component with related code, to make it clear. it was there initially for demo purpuses, and now, since we have home page, it's not needed imo

olenkabilonizhka: Thanks. I solved it. As a result, I got a lint error, because after cleaning the app component, we don't use users slice anywhere. Should we leave it?

PR #28, opened 2025-07-22
feat: add sign up backend flow cy-5

astowneer:


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,45 @@
...
+		const { email, name } = userRequestDto;
+
+		const existingUserByName = await this.userService.findByName(name);
+
+		if (existingUserByName) {
+			throw new AuthorizationError({
+				message: UserValidationMessage.NAME_ALREADY_EXISTS,
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
```
daniil-tkalenko, line 31: I'm not sure we should restrict users from having a same name, but it is more to QA team

packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,6 +1,6 @@
 const AuthApiPath = {
 	ROOT: "/",
-	SIGN_UP: "/sign-up",
+	SIGN_UP: "/register",
```
daniil-tkalenko, line 3: why did you decide to change the path?

apps/backend/src/libs/modules/encryption/encryption.ts
```diff
@@ -0,0 +1,8 @@
+import { config } from "~/libs/modules/config/config.js";
+
+import { BaseEncryption } from "./base-encryption.module.js";
+
+const encrypt = new BaseEncryption(config);
```
WilhelmWesser, line 5: Let's not load the whole config but pass only necessary arguments to the constructor in a manner like `{salt: SALT_FROM_ENV, etc.}`

apps/backend/src/libs/modules/encryption/libs/types/encrypt.type.ts
```diff
@@ -0,0 +1,6 @@
+type Encryption = {
```
WilhelmWesser, line 1: Let's rename to `Encryptor`

apps/backend/src/modules/users/user.service.ts
```diff
@@ -9,21 +10,25 @@ import {
 } from "./libs/types/types.js";
 
 class UserService implements Service {
+	private encryption: Encryption;
```
WilhelmWesser, line 13: encryptor

packages/shared/src/modules/users/libs/enums/user-validation-regex-rule.enum.ts
```diff
@@ -0,0 +1,15 @@
+import { UserValidationRule } from "./user-validation-rule.enum.js";
+
+const UserValidationRegexRule = {
+	EMAIL_VALID_CHARS_MIN_MAX: new RegExp(
```
WilhelmWesser, line 4: Let's make sure these regex conditions are aligner with QA team so that they will not forget to test it

WilhelmWesser (changes requested):

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,36 @@
+import { UserValidationMessage } from "shared/src/modules/users/libs/enums/user-validation-message.enum.js";
```
WilhelmWesser, line 1: Do not directly import anything from shared. Add this enum to the apps/backend/src/modules/auth/libs/enums/enums.ts file. Import that type there and export it so that you can import the enum into this service. Just like you do with AuthorizationError

WilhelmWesser (changes requested):

apps/backend/src/libs/modules/encryptor/base-encryptor.module.ts
```diff
@@ -0,0 +1,57 @@
...
+		const storedHashBuffer = Buffer.from(storedHash, "hex");
+
+		if (computedHash.length !== storedHashBuffer.length) {
+			return false;
+		}
+
+		return timingSafeEqual(computedHash, storedHashBuffer);
+	}
+
+	public async encrypt(value: string): Promise<{ hash: string; salt: string }> {
```
WilhelmWesser, line 41: Let's move { hash: string; salt: string } to a seprate type

apps/backend/src/libs/modules/encryptor/base-encryptor.module.ts
```diff
@@ -0,0 +1,57 @@
...
+		const salt = this.generateSalt();
+		const hashBuffer = await this.generateHash(value, salt);
+
+		return { hash: hashBuffer.toString("hex"), salt };
+	}
+
+	private async generateHash(value: string, salt: string): Promise<Buffer> {
+		return (await scryptAsync(value, salt, this.keyLength)) as Buffer;
```
WilhelmWesser, line 49: Does not scryptAsync return Buffer? If not cast to Buffer is ok

apps/backend/src/libs/modules/encryptor/base-encryptor.module.ts
```diff
@@ -0,0 +1,57 @@
...
+		this.saltSize = saltSize;
+	}
+
+	public async compare(
+		value: string,
+		storedHash: string,
+		salt: string,
+	): Promise<boolean> {
+		const computedHash = await this.generateHash(value, salt);
+		const storedHashBuffer = Buffer.from(storedHash, "hex");
```
WilhelmWesser, line 32: move "hex" to local constant

WilhelmWesser (changes requested):

apps/backend/src/libs/modules/encryptor/base-encryptor.module.ts
astowneer: It returns Promise<unknown>

WilhelmWesser: Could not we then simply do `return scryptAsync(value, salt, this.keyLength)`?

apps/backend/src/db/migrations/20250723203420_remove_unique_username.ts
```diff
@@ -0,0 +1,18 @@
...
+	});
+}
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dropUnique([COLUMN_NAME]);
+	});
+}
+
+export { down, up };
```
roman-mykytka, line 18: If you removed the name uniqueness check [here](https://github.com/BinaryStudioAcademy/bsa-2025-checkly/pull/28#discussion_r2226650835), please update it here as well to maintain consistency. I believe this is no longer needed anyway, since the uniqueness constraint was intentionally removed.

roman-mykytka (changes requested):

roman-mykytka (approved):

apps/backend/src/libs/modules/encryptor/base-encryptor.module.ts
astowneer: Currently, scrypt is used with promisify, which adds complexity and requires casting the result to Buffer. Alternatively, we could use the callback-based version of scrypt, but that introduces even more code due to callback handling. We can simplify by using scryptSync, but we loose async benefits. I think current version is optimal one with scrypt.

Issue #9, opened 2025-07-18
Protected route


daniil-tkalenko:
Implement authorization plugin for endpoints. 
- It should check signed in users by JWT and add inject it to request.
- Add possibility to exclude some endpoints - whiteList with rotes sign-up/sign-in

brunoazvd:
I have run all the backend protected routes test cases, including valid, invalid, and expired JWTs,  whitelisted routes and non-existing routes. Everything is working as expected after the Swagger route fix.  

Closing the ticket now.

PR #29, opened 2025-07-22
feat(backend): implement jwt authorization plugin cy-9

Amicuchi:
### Description

This pull request introduces a robust JWT-based authorization plugin for the Fastify server, fulfilling the requirements of Task #9. The primary goal is to protect endpoints by verifying a signed-in user's JWT and injecting their data into the request object for use in subsequent handlers.

### Key Implementations

- **Fastify Authorization Plugin:** A modular and reusable plugin (`authorization.plugin.ts`) was created to encapsulate all authentication and authorization logic. It leverages Fastify's hook system (`onRequest`) to intercept all incoming requests.

- **JWT Strategy:** The core token validation logic is handled by a dedicated strategy (`jwt.strategy.ts`) using the `jose` library. It is responsible for:
    - Extracting the `Bearer` token from the `Authorization` header.
    - Verifying the token's signature and validity against the `JWT_SECRET_KEY`.
    - Fetching the corresponding user from the database via `userService.findById()` to ensure the user exists.

- **Request Decoration:** The Fastify instance is decorated with an `authenticate` method, and the `request` object is augmented with a `user` property (`request.user`). This provides a clean and strongly-typed API for protected routes to access authenticated user data.

- **Route Whitelisting:** A mechanism to exclude public endpoints from authentication has been implemented. By adding an `isPublic: true` flag to a route's definition in the controller, the authorization plugin will automatically bypass the JWT check for that specific route. This was applied to the `/auth/sign-up` endpoint.

### Technical Decisions

- **`fastify-plugin`:** Used to ensure proper encapsulation and prevent context conflicts, following Fastify's best practices.
- **`jose`:** Chosen as the JWT library for its modern, promise-based API and robust security features.
- **Dependency Injection:** The `userService` is passed into the plugin during registration, promoting loose coupling and making the plugin more testable and reusable.


apps/backend/src/plugins/authorization/libs/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,43 @@
...
+import { AuthorizationError } from "../../exceptions/exceptions.js";
+
+const verifyJwt = async (request: FastifyRequest): Promise<UserEntity> => {
+	try {
+		const { authorization } = request.headers;
+
+		if (!authorization?.startsWith("Bearer ")) {
+			throw new AuthorizationError({
+				message: "Authorization header is missing or invalid.",
+			});
```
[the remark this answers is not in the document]
Amicuchi: All hardcoded error strings have been moved to a shared ErrorMessage enum to ensure consistency and reusability across the application.

apps/backend/src/plugins/authorization/libs/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,43 @@
...
+			new TextEncoder().encode(config.ENV.JWT.SECRET_KEY),
+		);
+
+		const { userService } = request.server;
+		const user = await userService.findById(payload.id);
+
+		if (!user) {
+			throw new AuthorizationError({
+				message: "User not found.",
+			});
```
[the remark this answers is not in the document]
Amicuchi: All hardcoded error strings have been moved to a shared ErrorMessage enum to ensure consistency and reusability across the application.

apps/backend/src/plugins/authorization/libs/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		return user;
+	} catch (error) {
+		throw new AuthorizationError({
+			cause: error,
+			message: "Authentication failed.",
```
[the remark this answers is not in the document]
Amicuchi: All hardcoded error strings have been moved to a shared ErrorMessage enum to ensure consistency and reusability across the application.

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -191,6 +193,15 @@ class BaseServerApplication implements ServerApplication {
 		);
 	}
 
+	private async initPlugins(): Promise<void> {
+		await this.app.register(authorizationPlugin, {
+			userService,
+			whiteRoutes: this.apis.flatMap((api) =>
+				api.routes.filter((route) => route.isPublic).map((route) => route.path),
+			),
```
daniil-tkalenko, line 201: consider moving it to separate private method


```
private getWhiteRoutes(): string[] {
    return this.apis.flatMap((api) =>
        api.routes
            .filter((route) => route.isPublic)
            .map((route) => route.path)
    );
}
```

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+const authorization = fp<AuthPluginOptions>(
+	(fastify, { userService, whiteRoutes }, done) => {
+		fastify.decorate("userService", userService);
+		fastify.decorate("authenticate", async function (request) {
+			request.user = await verifyJwt(request);
+		});
+
+		fastify.addHook("onRequest", async (request, reply) => {
```
daniil-tkalenko, line 33:
```suggestion
		fastify.addHook("preHandler", async (request, reply) => {
```

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+			const routeUrl = request.routeOptions.url ?? "";
```
daniil-tkalenko, line 34:
```suggestion
			const routeUrl = request.routerPath || request.routeOptions.url;
```

apps/backend/src/plugins/authorization/libs/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,46 @@
+import { type FastifyRequest } from "fastify";
+import { type JWTPayload, jwtVerify } from "jose";
+import { AuthorizationError, ErrorMessage } from "shared";
+
+import { config } from "../../../../libs/modules/config/config.js";
+import { type UserEntity } from "../../../../modules/users/user.entity.js";
```
daniil-tkalenko, line 6:
```suggestion
import { config } from "~/libs/modules/config/config.js";
import { type UserEntity } from "~/modules/users/user.entity.js";
```

apps/backend/src/plugins/authorization/libs/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,46 @@
...
+			throw new AuthorizationError({
+				message: ErrorMessage.AUTHORIZATION_HEADER_MISSING,
+			});
+		}
+
+		const token = authorization.replace("Bearer ", "");
+		const { payload } = await jwtVerify<JWTPayload & { id: number }>(
+			token,
+			new TextEncoder().encode(config.ENV.JWT.SECRET_KEY),
+		);
```
daniil-tkalenko, line 26: once https://github.com/BinaryStudioAcademy/bsa-2025-checkly/pull/35 is merged, please use created token module here to decode the token

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+	whiteRoutes: string[];
+};
+
+const authorization = fp<AuthPluginOptions>(
+	(fastify, { userService, whiteRoutes }, done) => {
+		fastify.decorate("userService", userService);
+		fastify.decorate("authenticate", async function (request) {
+			request.user = await verifyJwt(request);
+		});
+
```
daniil-tkalenko, line 32: I think we do not really need this, since we have userService in props already. fastify.decorate could be usefull if we are going to reuse this service on other routes, or it's something global

if it's just to pass to verifyJwt helper, I would propose to pass it as props OR even move logic from verifyJwt to plugin, and handle all needed actions here. there is not many logic, try to make it neat in the scope of plugin.

daniil-tkalenko (changes requested):

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -75,6 +75,7 @@ class AuthController extends BaseController {
 						body: UserSignUpRequestDto;
 					}>,
 				),
+			isPublic: true,
```
daniil-tkalenko, line 78: Sign in above should be public too

apps/backend/src/plugins/authorization/authorization.plugin.ts
Amicuchi: I've used a differente approach.
const routeUrl = request.routeOptions.url || request.url;
when I tried to use the "request.routerPath" got an error, saying that this is not a valid property of Fatify, so I've founded this one.

apps/backend/src/libs/modules/server-application/base-server-application.ts
Amicuchi: done!

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,13 +1,14 @@
...
-import { UserEntity } from "~/modules/users/user.entity.js";
-import { type UserRepository } from "~/modules/users/user.repository.js";
-
 import {
+	type UserGetAllItemResponseDto,
 	type UserGetAllResponseDto,
 	type UserSignUpRequestDto,
 	type UserSignUpResponseDto,
-} from "./libs/types/types.js";
+} from "shared";
```
daniil-tkalenko, line 6: here you import these types from shared directly. it's incorrect approach - when you need something from shared, create libs folder alongside with your files and add there re-exports from shared. then in your file you can import types or others things from the nearest libs, not shared directly.
here prev version of code is actioally how it should be done, and also libs/types folder already present, check it as a reference

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,52 @@
+import { type FastifyReply, type FastifyRequest } from "fastify";
+import fp from "fastify-plugin";
+import micromatch from "micromatch";
+import { type UserSignUpResponseDto } from "shared";
```
daniil-tkalenko, line 4: same here

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,52 @@
...
+type AuthPluginOptions = {
+	userService: UserService;
+	whiteRoutes: string[];
+};
+
+const authorization = fp<AuthPluginOptions>(
+	(fastify, { userService, whiteRoutes }, done) => {
+		fastify.decorate("authenticate", async function (request: FastifyRequest) {
+			request.user = await verifyJwt(request, userService);
+		});
```
daniil-tkalenko, line 29: I would also move logic from verifyJwt to the plugin - not sure we would use authenticate function elsewhere, so it will be cleaner and easier to have the logic on one place. but if you want to keep it - rename verifyJwt to extractUserFromRequest - it will be more descriptive name for the function

daniil-tkalenko (changes requested):

Amicuchi:
- The JWT verification logic was moved to a standalone `extractUserFromRequest` function 
- All imports from "shared" are now correctly routed through the local `libs/types.ts` file as per project guidelines.
- The `preHandler` hook signature was cleaned up by removing the unused `reply` parameter.
- Type definitions for Fastify were added to ensure properties like `request.user` are correctly recognized by TypeScript.

apps/backend/package.json
```diff
@@ -28,6 +28,9 @@
 		"convict": "6.2.4",
 		"dotenv": "17.2.0",
 		"fastify": "5.4.0",
+		"fastify-plugin": "5.0.1",
+		"jose": "6.0.12",
+		"micromatch": "^4.0.5",
```
WilhelmWesser, line 33: The version should be fixed (without the carret)

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,90 @@
...
+		if (!authorization?.startsWith("Bearer ")) {
+			throw new AuthorizationError({
+				message: ErrorMessage.AUTHORIZATION_HEADER_MISSING,
+			});
+		}
+
+		const token = authorization.replace("Bearer ", "");
+
+		// TODO: once #35 is merged, must use created token module here to decode the token
+		const { payload } = await jwtVerify<JWTPayload & { id: number }>(
```
WilhelmWesser, line 46: Let's move JWTPayload & { id: number } to a separate type

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,90 @@
...
+	try {
+		const { authorization } = request.headers;
+
+		if (!authorization?.startsWith("Bearer ")) {
+			throw new AuthorizationError({
+				message: ErrorMessage.AUTHORIZATION_HEADER_MISSING,
+			});
+		}
+
+		const token = authorization.replace("Bearer ", "");
```
WilhelmWesser, line 43: Please, move "Bearer" to a separate constant

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+const extractUserFromRequest = async (
+	request: FastifyRequest,
+	userService: UserService,
+): Promise<UserSignUpResponseDto> => {
+	try {
+		const { authorization } = request.headers;
+
+		if (!authorization?.startsWith("Bearer ")) {
```
WilhelmWesser, line 37: Please, move "Bearer" to a separate constant and reuse everywhere where needed this constant instead of a string

WilhelmWesser (approved):

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		const token = authorization.replace(AuthStrategy, "");
+
+		const { payload } = await jwtVerify<JWTPayloadWithId>(
+			token,
+			new TextEncoder().encode(config.ENV.JWT.SECRET_KEY),
+		);
+
+		const user = await userService.findById(payload.id);
```
daniil-tkalenko, line 54: we have merged task with JWT generation - please, resolve conflicts, use here token module to decode the token, and I belieave we can merge this PR next

daniil-tkalenko (changes requested):

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+			});
+		}
+
+		const tokenValue = authorization.replace(AuthStrategy, "");
+
+		const payload = (await token.decodeToken(tokenValue)) as { userId: number };
+
+		const user = (await userService.findById(
+			payload.userId,
+		)) as null | UserEntity;
```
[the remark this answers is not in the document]
Amicuchi: I don't know why, I assumed that the findById service method was returning a full UserEntity instance and not a plain DTO. 
And the "as null | UserEntity" type assertion to force TypeScript to accept what I thought was the correct type, which then led me to incorrectly call .toObject() on it.
Calling .toObject() would indeed cause a runtime error. 
I have removed both the unnecessary type assertion and the incorrect .toObject() call.

WilhelmWesser (approved):

PR #35, opened 2025-07-22
Feature/jws token

nobiqp:
-Implemented JWT token generation on the back end and included it in the response payload upon successful authentication.
-Added a helper utility on the front end to manage the token: store it in localStorage, retrieve it for authenticated requests, and delete it on sign-out.


apps/backend/src/modules/auth/jwt-auth/auth.jwt.ts
```diff
@@ -0,0 +1,22 @@
...
+		.setProtectedHeader({ alg: "HS256" })
+		.setSubject(userId)
+		.setIssuedAt()
+		.setExpirationTime("1d")
+		.sign(secret);
+
+	return token;
+}
+
+export { generateJWT };
```
WilhelmWesser, line 22: Also your service must include a method to decod JWT token to get the user data out of there

WilhelmWesser (changes requested):

apps/backend/src/libs/modules/encrypt/encrypt.ts
```diff
@@ -0,0 +1,6 @@
+import { BaseEncrypt } from "./base-encrypt.module.js";
+
+const encrypt = new BaseEncrypt();
```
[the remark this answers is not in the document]
nobiqp: This part of the code and a few others from sign up back end, I just merged it to my JWT branch so I have some back end to rely on. Thanks you for all feedback!

apps/frontend/src/pages/auth/auth-token.helper.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const getToken = (): null | string => {
+	return localStorage.getItem("token");
+};
+
+const clearToken = (): void => {
+	localStorage.removeItem("token");
+};
+
+export {clearToken, getToken, storeToken}
```
roman-mykytka, line 13: extract into a dedicated storage service, similar to the encrypt service

nobiqp: @roman-mykytka @WilhelmWesser 

Thank you so much for your feedback! Currently working on moving it to dedicated service.

nobiqp:
-Created service for JWT on back end to generate, verify and decode a token
-Now token properly transferred from back end to front end
-Returned dotenv
-Installed jose to the backend only

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -188,13 +190,20 @@ class AuthController extends BaseController {
...
 	): Promise<APIHandlerResponse> {
+		const user = await this.authService.signUp(options.body);
+
+		user.token = await this.jwtService.generateToken(user.id.toString());
+
 		return {
-			payload: await this.authService.signUp(options.body),
+			payload: {
+				user,
+			},
```
daniil-tkalenko, line 206: you should remove this changes from controller, and instead add it to service
service should return object like this: { user, token }, no need to place token inside of user

apps/frontend/src/modules/auth/token-storage/token-storage.service.ts
```diff
@@ -0,0 +1,19 @@
...
+	public get(): null | string {
+		return localStorage.getItem(this.tokenKey);
+	}
+
+	public store(token: string): void {
+		localStorage.setItem(this.tokenKey, token);
+	}
+}
+
+export { BaseTokenStorage };
```
daniil-tkalenko, line 19: we already have BaseStorage module with the same functionality
why do we need specific BaseTokenStorage? if no resone, please remove it and use BaseStorage

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -30,20 +33,38 @@ const Auth: React.FC = () => {
...
+			const isSignUpFulfilled =
+				authActions.signUp.fulfilled.match(resultAction);
+
+			if (isSignUpFulfilled) {
+				const { token } = resultAction.payload;
+
+				if (token) {
+					tokenStorage.store(token);
+				}
+			}
```
daniil-tkalenko, line 48: here you should only dispatch the action
logic that set token to storage should be added in signUp/signIn  createAsyncThunk actions

note: update ExtraArguments to contain storage as well as authApi and userApi
if will allow you to have access to storage  in the actions like this: const { authApi, storage } = extra;

daniil-tkalenko (changes requested):

knip.config.ts
```diff
@@ -1,6 +1,8 @@
 import { type KnipConfig } from "knip";
 
 const config: KnipConfig = {
+	ignore: ["apps/frontend/src/pages/auth/auth-token.helper.ts"],
```
WilhelmWesser, line 4: You should not add any files to ignore but fix the errors

WilhelmWesser (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,48 @@
...
+		const { payload } = await jwtVerify(token, this.secret);
+
+		return {
+			expirationTime: new Date(
+				(payload.exp ?? DEFAULT_TIMESTAMP_SECONDS) * SECONDS_TO_MILLISECONDS,
+			).toISOString(),
+			issuedAt: new Date(
+				(payload.iat ?? DEFAULT_TIMESTAMP_SECONDS) * SECONDS_TO_MILLISECONDS,
+			).toISOString(),
+			userId: Number(payload.sub),
```
daniil-tkalenko, line 35: we changed generateToken to accept userId as payload, and removed subject, so I think sub will not work here anymore. We should extract userId like this, or add it to subject as you did initially - up to you 
```suggestion
			userId: Number(payload['userId']),
```

PR #46, opened 2025-07-25
feat: add home page wrapper cy-43

watchlar2000:
Hello!

Here's the list of changes:
- home component lives inside pages folder. All other sections, header and footer that will be built later should go inside pages/home/components/
```
├──  pages
│  ├──  home
│  │  ├──  components
│  │  │  ├──  header
│  │  │  ├──  nav
│  │  │  └──  components.ts
│  │  ├──  home.tsx
│  │  └──  styles.module.css
│  └──  pages.ts
```

- home component is added to the router provider as a public page
- inside home component, a few sections are mocked for demo purposes

Demo:

image

### DIFF
--- a/apps/frontend/src/pages/home/components/nav/nav.tsx
+++ b/apps/frontend/src/pages/home/components/nav/nav.tsx
@@ -1,5 +1,3 @@
-import React from "react";
-
 import { Link } from "~/libs/components/components.js";
 
 const Nav: React.FC = () => {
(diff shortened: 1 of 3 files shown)


daniil-tkalenko (approved):

apps/frontend/src/pages/home/components/nav/nav.tsx
```diff
@@ -0,0 +1,14 @@
+import { type JSX } from "react";
+
+import { Link } from "~/libs/components/components.js";
+
+const Nav = (): JSX.Element => {
```
WilhelmWesser, line 5: let's type Nav as React.FC and remove JSX.Element return type. Make similar changes in the whole your PR

WilhelmWesser (changes requested):

PR #47, opened 2025-07-25
feat: Sign up flow added cy-3

Volin13:
add validation schema to shared

implemented all action statuses (pending, fulfilled, rejected), add sign up user to store on success + added error handlers

sign in link navigate to sign in page

also added some images styles and base reset styles


apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -1,56 +1,125 @@
...
+	const { control, errors, handleSubmit } =
+		useAppForm<SignUpFormValidationSchema>({
+			defaultValues: { ...DEFAULT_SIGN_UP_PAYLOAD, confirmPassword: "" },
+			validationSchema: userSignUpValidationSchemaExtended,
+		});
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
-			void handleSubmit(onSubmit)(event_);
+			void handleSubmit((data) => {
+				// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
daniil-tkalenko, line 34:
```suggestion
				const { confirmPassword: _, ...payloadToSend } = data;
```
it tells linter that var can be unused

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -10,26 +10,59 @@ import {
...
+				state.error = {
+					message: action.error.message,
+					status: 500,
+				};
+			} else {
+				state.error = {
+					message: "Unknown error! Try again later.",
+					status: 500,
+				};
+			}
```
daniil-tkalenko, line 65: for now I think you can just use showErrorToast(action.error.message as string); as done for sign in
proper notifications will be done in other tasks that already in progress

daniil-tkalenko (changes requested):

apps/frontend/src/assets/img/sign-up/yellow-stars.png
WilhelmWesser, line 1: There yellow stars image asset is already present in sign-in folder. There is no need to add a new one

apps/frontend/src/assets/img/sign-up/blue-stars.png
WilhelmWesser, line 1: There blue stars image asset is already present in sign-in folder. There is no need to add a new one

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -10,26 +10,59 @@ import {
...
 import { signIn, signUp } from "./actions.js";
 
+interface BackendErrorResponse {
+	message: string;
+	status: number;
+}
+
 type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
+	error: BackendErrorResponse | null;
```
WilhelmWesser, line 20: there is no need in storing error in the state. When error appears we just notify user about it and that's it. Remove it from here

apps/frontend/src/modules/users/libs/validation-schemas/validation-schemas.ts
```diff
@@ -1 +1,5 @@
-export { userSignInValidationSchema, userSignUpValidationSchema } from "shared";
+export {
+	type SignUpFormValidationSchema,
```
WilhelmWesser, line 2: you should import types from shared folder in types.ts file and import it from there

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -1,56 +1,125 @@
...
+							placeholder="********"
+							required
+							type="password"
+						/>
+						<Button label="Create an account" type="submit" />
+					</form>
+				</div>
+				<img
+					alt="blue stars"
+					className={`${css["image-position"] as string} ${css["blue-stars"] as string}`}
```
WilhelmWesser, line 103: Use `getClassNames()` for classes concatenation

WilhelmWesser (changes requested):

apps/frontend/src/assets/img/sign-up/blue-stars.png
Volin13: done

apps/frontend/src/modules/auth/slices/auth.slice.ts
Volin13: done

apps/frontend/src/modules/auth/slices/auth.slice.ts
Volin13: done

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -1,56 +1,125 @@
...
+			validationSchema: userSignUpValidationSchemaExtended,
+		});
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
-			void handleSubmit(onSubmit)(event_);
+			void handleSubmit((data) => {
+				const payloadToSend = { ...data };
+				delete (payloadToSend as Partial<SignUpFormValidationSchema>)
+					.confirmPassword;
+				onSubmit(payloadToSend as UserSignUpRequestDto);
```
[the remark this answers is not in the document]
Volin13: This case doesn`t cause lint errror but idk if it is goo idea to do it like that. Right now my data (that i got from my form) is
data: {
    email: string;
    name: string;
    password: string;
    confirmPassword: string;
    
    so it has confirmPassword key that is not need in the backend (as far as I know). My backend uses type UserSignUpRequestDto = {
    email: string;
    name: string;
    password: string;
}
If I will send payLoad with confirmPassword it would cause error at the backend (maybe?). I could change like this UserSignUpRequestDto = {
    email: string;
    name: string;
    password: string;
    confirmPassword?: string;
} but I still dont know if it is ok to send confirmPassword that is not needed at the back end. If you want we could discuss this issue in slack call or mb here

daniil-tkalenko (approved):

WilhelmWesser (approved):

PR #66, opened 2025-07-28
feat: Refactor the way notifications (toasts) cy-45

WhoeverWhenever:
- Added a toast-notifier service
- Added a listener middleware to show a toast on rejection


apps/frontend/src/libs/modules/toast-notifier/toast-notifier.module.ts
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	public info(message: string): string {
+		return toast.info(message, this.toastOptions).toString();
+	}
+
+	public success(message: string): string {
+		return toast.success(message, this.toastOptions).toString();
+	}
+}
```
roman-mykytka, line 27: - prefer a more generic class name—Notification
- expose the default options as read‑only to avoid accidental mutation.
- when callers need custom behaviour, let them pass an options object to the method rather than modifying the defaults.

roman-mykytka (changes requested):

apps/frontend/src/libs/modules/notifications/notifications.module.ts
```diff
@@ -0,0 +1,39 @@
...
+	): string {
+		return toast.info(message, notificationOptions).toString();
+	}
+
+	public success(
+		message: string,
+		notificationOptions: ToastOptions = this.defaultNotificationOptions,
+	): string {
+		return toast.success(message, notificationOptions).toString();
+	}
```
WhoeverWhenever, line 36: Is it better to do like this:
```
public success(
		message: string,
		notificationOptions: ToastOptions = this.defaultNotificationOptions,
	): string {
		const customNotificationOptions = {
			...this.defaultNotificationOptions,
			...notificationOptions,
		};

		return toast.success(message, customNotificationOptions).toString();
	}
```
so users can pass parameters to update defaults and have no need to go through all the other options:
`const newOptions: ToastOptions = { transition: Bounce };`

roman-mykytka: Yes — merging the override with the defaults inside the method is the safer, more ergonomic approach.

Also, consider adding a private options() helper method to avoid repeating the spread logic in every notification method:

`private static options(overrides?: Partial<ToastOptions>): ToastOptions {
  return { ...this.defaultNotificationOptions, ...overrides };
}
`

roman-mykytka (changes requested):

apps/frontend/src/libs/modules/notifications/notifications.module.ts
```diff
@@ -0,0 +1,52 @@
...
+	private static defaultNotificationOptions: Readonly<ToastOptions> = {
+		autoClose: NotificationOptionValues.AUTO_CLOSE_TIME,
+		closeOnClick: false,
+		draggable: true,
+		hideProgressBar: true,
+		pauseOnHover: true,
+		position: "top-center",
+		theme: "colored",
+		transition: Zoom,
+	};
```
WilhelmWesser, line 15: This can be moved out of the class at all and placed as a constant

WilhelmWesser (changes requested):

WilhelmWesser (approved):

apps/frontend/src/libs/modules/notifications/libs/enums/notification-option-values.enum.ts
```diff
@@ -0,0 +1,5 @@
+const NotificationOptionValues = {
```
WilhelmWesser, line 1: Seems that NotificationOptionValues is not an enum actually, so there is no need to define an object. Move AUTO_CLOSE_TIME out just like a common constant value

WilhelmWesser: Also move it to a constant file

WilhelmWesser (changes requested):

apps/frontend/src/libs/modules/store/listener-middleware/listener-middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+
+listenerMiddleware.startListening({
+	effect: (action) => {
+		const { message } = action.error;
+
+		if (message?.trim()) {
+			notifications.error(action.error.message as string);
+		} else {
+			notifications.error(ErrorMessage.DEFAULT_ERROR_MESSAGE);
+		}
```
daniil-tkalenko, line 16:
```suggestion
		const { message } = action.error ?? {};

		if (message?.trim()) {
			notifications.error(message);
		} else {
			notifications.error(ErrorMessage.DEFAULT_ERROR_MESSAGE);
		}
```

apps/frontend/src/modules/users/slices/actions.ts
```diff
@@ -3,16 +3,20 @@ import { createAsyncThunk } from "@reduxjs/toolkit";
...
 	undefined,
 	AsyncThunkConfig
->(`${sliceName}/load-all`, (_, { extra }) => {
-	const { userApi } = extra;
+>(`${sliceName}/load-all`, async (_, { extra }) => {
+	const { notifications, userApi } = extra;
+	const result = await userApi.getAll();
 
-	return userApi.getAll();
+	notifications.success(UserNotificationMessages.GET_ALL_USERS_SUCCESS);
```
daniil-tkalenko, line 17: I would remove this toast, this is test request and it fires on root page as I remember, so not sure we need it

PR #70, opened 2025-07-28
fix: fix-sign-in-backend cy-69

muxkuro:
Added middleware which checks if error has status code for AuthorizationError, or if this error has details option like ValidationError and maps it to the same structure. 
Added a hook onRequest sent, which throws it in the middleware. Only after middleware is done checking and restructuring the error, the request is sent.

Fixed some validation rules I originally used, which resulted in a bug.


apps/backend/src/libs/modules/middleware/error-mapper.middleware.ts
```diff
@@ -0,0 +1,69 @@
...
+				payload &&
+				typeof payload === "object"
+			) {
+				const errorPayload = payload as ErrorPayload;
+
+				let errorType: string = ServerErrorType.COMMON;
+				let message: string = errorPayload.message || DEFAULT_ERROR_MESSAGE;
+
+				if (
+					errorPayload.details &&
```
WilhelmWesser, line 45: Here also `errorPayload.details` condition looks redundant because Array.isArray(errorPayload.details) will return false in case errorPayload.details is falsy

apps/backend/src/libs/modules/middleware/error-mapper.middleware.ts
```diff
@@ -0,0 +1,69 @@
...
+					Array.isArray(errorPayload.details) &&
+					errorPayload.details.length >= MIN_DETAILS_LENGTH
+				) {
+					errorType = ServerErrorType.VALIDATION;
+					message =
+						errorPayload.details[FIRST_DETAIL_INDEX]?.message ||
```
WilhelmWesser, line 51: Let's be more specific. Here, we expect to get an error message from different objects. We use || operator to get an error message from another object in case it is absent in the current one. Let's replace || with the ?? operator. It will go to the alternative object (the one that is to the right of the ?? operator) only in case the value from left is undefined or null (generally absent). Replace it in all related places in terms of this PR

apps/backend/src/libs/modules/middleware/error-mapper.middleware.ts
```diff
@@ -0,0 +1,69 @@
...
+						errorPayload.message ||
+						DEFAULT_ERROR_MESSAGE;
+				}
+
+				const standardizedError = {
```
WilhelmWesser, line 56: Can we type it as "GeneralError"?

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -191,6 +194,10 @@ class BaseServerApplication implements ServerApplication {
 		);
 	}
 
+	private initErrorMapperMiddleware(): void {
+		this.app.addHook("onRequest", initErrorMapperMiddleware());
```
WilhelmWesser, line 198: Please, move the "onRequest" to an enum

apps/backend/src/libs/modules/middleware/error-mapper.middleware.ts
roman-mykytka: additionally

I prefer using `const` instead of `let` to avoid unnecessary `if` statements and variable reassignment. In this case, you can extract the condition into a separate `hasDetails` variable and then define both `errorType` and `message` as `const` using ternary expressions. This makes the logic clearer, more concise, and easier to maintain.

like:
```
const hasDetails =
  Array.isArray(errorPayload.details) &&
  errorPayload.details.length >= MIN_DETAILS_LENGTH;

const errorType: ServerErrorType = hasDetails
  ? ServerErrorType.VALIDATION
  : ServerErrorType.COMMON;

const message: string =
  (hasDetails
    ? errorPayload.details[FIRST_DETAIL_INDEX]?.message
    : errorPayload.message) ?? DEFAULT_ERROR_MESSAGE;
```

roman-mykytka (changes requested):

apps/backend/src/libs/modules/middleware/error-mapper.middleware.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type FastifyPayload =
+	| boolean
+	| ErrorPayload
+	| null
+	| number
+	| Record<string, unknown>
+	| string
+	| undefined
+	| unknown[];
```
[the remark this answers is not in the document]
muxkuro: you're right. I was trying to understand what types could fastify payload return and tried to make union. I decided to just delete this type at all, since we can just use unknown next to the argument in the function.

daniil-tkalenko (approved):

PR #78, opened 2025-07-29
feat: add input and button components cy-14

watchlar2000:


apps/frontend/package.json
```diff
@@ -35,6 +35,7 @@
 		"eslint-plugin-jsx-a11y": "6.10.2",
 		"eslint-plugin-react": "7.37.5",
 		"eslint-plugin-react-hooks": "5.2.0",
-		"vite": "7.0.4"
+		"vite": "7.0.4",
```
WilhelmWesser, line 38: we already have vite installed in frontend devDeps, remove it from here

apps/frontend/src/pages/home/components/hero/hero.tsx
```diff
@@ -5,64 +5,71 @@ import {
...
-						styles["pink-stars"],
-					)}
-					src={pinkStars}
-				/>
-				<img
-					alt=""
-					className={getClassNames(styles["floating-image"], styles["cat"])}
-					src={cat}
-				/>
+const decorativeImages = [
```
WilhelmWesser, line 14: Make is a constant list and type it

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -27,23 +31,65 @@ const Input = <T extends FieldValues>({
 	required,
 	type = "text",
 }: Properties<T>): JSX.Element => {
+	const [showPassword, setShowPassword] = useState<boolean>(false);
+	const inputId = useId();
 	const { field } = useFormController({ control, name });
-
 	const error = errors[name]?.message;
 	const hasError = Boolean(error);
```
daniil-tkalenko, line 38:
```suggestion
	const errorMessage = errors[name]?.message;
```

apps/frontend/package.json
watchlar2000: I didn't install vite, it just shows that comma was added after it, because another lib was added to the devDeps - svgr plugin

apps/frontend/src/libs/components/logo/logo.tsx
```diff
@@ -1,14 +1,20 @@
...
 
 import styles from "./styles.module.css";
 
 const Logo: React.FC = () => {
 	return (
-		<div className={getClassNames(styles["container"], "cluster")}>
-			<img alt="Logo icon" src={logoIcon} />
-			<h1>Checkly</h1>
-		</div>
+		<Link
```
watchlar2000, line 11: @WilhelmWesser @daniil-tkalenko guys, I've also updated the Logo component a bit - wrapped it with Link component so logo navigates to root page when clicked. Is that ok or keep it wrapped with DIV and wrap it with link inside the component that imports this Logo component?

apps/frontend/package.json
```diff
@@ -35,6 +35,7 @@
...
+		"vite-plugin-svgr": "4.3.0"
```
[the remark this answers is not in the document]
WilhelmWesser: :fire:

roman-mykytka (approved):

PR #121, opened 2025-08-02
fix: authenticated user access control cy-110

olenkabilonizhka:
Updated ProtectedRoute functionality.

Now we have 3 access levels:
- for all.
- for authenticated users.
- for not authenticated users.


apps/frontend/src/index.tsx
```diff
@@ -11,39 +11,20 @@ import {
...
-								],
-								element: <ProtectedRoute />,
-								path: AppRoute.DASHBOARD,
-							},
-						],
-						element: <App />,
+						children: routes,
+						element: (
+							<App>
+								<ProtectedRoute />
```
WilhelmWesser, line 25: I feel that ProtectedRoute component should wrap pages that need authentication rather then being and intermidiary router. Do you agree?

WilhelmWesser (changes requested):

apps/frontend/src/index.tsx
olenkabilonizhka: I created it to fix the following bug: authenticated users can access the sign-in & sign-up pages. I think we need to handle routing access based on authentication status, some pages should be available only to authenticated users (dashboard, plan, ...), some only to not authenticated users (sign-in, sign-up, ...), and some should be accessible to everyone (home page with general info about the app, not found page). Based on these requirements I created this PR. Maybe ProtectedRoute should be renamed to something like AccessRoute to better reflect the logic.
Do you suggest renaming, splitting this into separate components or would you recommend another approach?
Personally, I find it easier to understand and support this functionality in one place.

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -1,15 +1,21 @@
...
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 import { useAuthInitialization } from "~/libs/hooks/use-auth-initialization/use-auth-initialization.hook.js";
+import { type RouteHandle } from "~/routes.js";
 
 import { Loader } from "../components.js";
 
 const ProtectedRoute: React.FC = () => {
 	useAuthInitialization();
 
+	const LAST_INDEX = -1;
```
roman-mykytka, line 14: move outside the component

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -24,11 +30,21 @@ const ProtectedRoute: React.FC = () => {
...
+	} else if (access === RouteAccess.AUTHENTICATED) {
+		if (isAuthorized) {
+			return <Outlet />;
+		}
 
-	return <Navigate replace to={AppRoute.SIGN_IN} />;
+		return <Navigate replace to={AppRoute.SIGN_IN} />;
+	} else {
+		return <Outlet />;
+	}
```
roman-mykytka, line 47: Deeply-nested if/else blocks quickly become hard to scan and reason about. A switch (or early returns) flattens the control flow, making the logic clearer.

```
switch (access) {
  case RouteAccess.NOT_AUTHENTICATED:
    return isAuthorized
      ? <Navigate replace to={AppRoute.DASHBOARD} />
      : <Outlet />;

  case RouteAccess.AUTHENTICATED:
    return isAuthorized
      ? <Outlet />
      : <Navigate replace to={AppRoute.SIGN_IN} />;

  default:
    return <Outlet />;
}

```

roman-mykytka (changes requested):

apps/frontend/src/index.tsx
olenkabilonizhka: I got it. I tried to implement. We can't simply exclude two pages, because they're outside the protected route and it doesn't get called. So, we need to wrap the entire App component in a protected route, which results in numerous conditional checks for proper navigation and hardcoded access levels to distinguish them. I think it might lead to redundant code and could be hard to scale. What do you think about it?

WilhelmWesser: Nope, I meant to wrap each page that requires the authentication with a ProtectedRoute component and do not wrap the pages that are accessible by not authenticated users

apps/frontend/src/libs/enums/route-access.ts
```diff
@@ -0,0 +1,7 @@
+const RouteAccess = {
+	ALL: "All",
```
WilhelmWesser, line 2: What is the difference between "All" and "Not authenticated" route accesses? I think we can remove "ALL" and use only "NOT_AUTHENTICATED"

WilhelmWesser: Also replace all places in the PR where ALL with NOT_AUTHENTICATED

WilhelmWesser (changes requested):

apps/frontend/src/libs/enums/route-access.ts
WilhelmWesser, line 2: Let's rename ALL to PUBLIC to make it more clear

WilhelmWesser (changes requested):

WilhelmWesser (approved):

PR #134, opened 2025-08-03
fix: update checkIsWhiteRoute function cy-132

watchlar2000:
Hi guys!

[Bug link](https://github.com/orgs/BinaryStudioAcademy/projects/33?pane=issue&itemId=122660075&issue=BinaryStudioAcademy%7Cbsa-2025-checkly%7C132)

The checkIsWhiteRoute function accepts two params - url (`'/api/v1/auth/login'`) and a list of white routes (like `['/api/v1/auth/login', '/api/v1/auth/register']`) but returns incorrect results. 

Issue is that the function extracted the route part (`'/auth/login'`) via regex and compared it with the full path in the whitelist `['/api/v1/auth/login', '/api/v1/auth/register']`, which would never be true.

Suggestion is to remove redundant regex extraction and just verify if white list includes the full url route (not including query params).

UPD: regex check is restored + replaced route.split("?") with url.split("?").

### DIFF
--- a/apps/backend/src/plugins/authorization/authorization.plugin.ts
+++ b/apps/backend/src/plugins/authorization/authorization.plugin.ts
@@ -75,9 +75,9 @@ const checkIsWhiteRoute = (url: string, whiteRoutes: string[]): boolean => {
 		return true;
 	}
 
-	const [routeWithoutQuery] = url.split("?") as [string];
+	const [routeWithoutQuery] = url.split("?");
 
-	return whiteRoutes.includes(routeWithoutQuery);
+	return whiteRoutes.includes(routeWithoutQuery as string);
 };
 
 const authorization = fp<AuthPluginOptions>(


WilhelmWesser:
> Hi guys!
> 
> [Bug link](https://github.com/orgs/BinaryStudioAcademy/projects/33?pane=issue&itemId=122660075&issue=BinaryStudioAcademy%7Cbsa-2025-checkly%7C132)
> 
> The checkIsWhiteRoute function accepts two params - url (`'/api/v1/auth/login'`) and a list of white routes (like `['/api/v1/auth/login', '/api/v1/auth/register']`) but returns incorrect results.
> 
> Issue is that the function extracted the route part (`'/auth/login'`) via regex and compared it with the full path in the whitelist `['/api/v1/auth/login', '/api/v1/auth/register']`, which would never be true.
> 
> Suggestion is to remove redundant regex extraction and just verify if white list includes the full url route (not including query params).

That is not a suggestion that would work because the goal of this regex is to validate all the routes that start with /api/v1/ (all our backend routes a expected to start like that) and in case it is not, we allow to route. It is done so because on deploy, our backend statically serves our frontend, and if we remove that regex and try to open the main page (our landing page), the request will be blocked. So the solution should not remove this regex but extend this logic. We can replace match() here with test() just to check if a route fits the whole regex. If no then return true if yes then check if this route is in the white list

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -76,9 +76,8 @@ const checkIsWhiteRoute = (url: string, whiteRoutes: string[]): boolean => {
 		return true;
 	}
 
-	const ROUTE_WITH_QUERY_INDEX = 0;
-
-	const routeWithoutQuery = route.split("?")[ROUTE_WITH_QUERY_INDEX] as string;
+	const URL_WITHOUT_QUERY_INDEX = 0;
+	const routeWithoutQuery = url.split("?")[URL_WITHOUT_QUERY_INDEX] as string;
```
roman-mykytka, line 80:
```suggestion
const [routeWithoutQuery] = url.split("?");
```

You can simplify this by using array destructuring instead of accessing the index with a constant. It improves readability and avoids unnecessary variable declarations.

roman-mykytka (changes requested):

apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -69,16 +69,13 @@ const extractUserFromRequest = async (
...
+	if (!isAPIRoute) {
 		return true;
 	}
 
-	const ROUTE_WITH_QUERY_INDEX = 0;
-
-	const routeWithoutQuery = route.split("?")[ROUTE_WITH_QUERY_INDEX] as string;
+	const [routeWithoutQuery] = url.split("?") as [string];
 
 	return whiteRoutes.includes(routeWithoutQuery);
```
WilhelmWesser, line 80:
```suggestion
	const [routeWithoutQuery] = route.split("?");

	return whiteRoutes.includes(routeWithoutQuery as string);
```
Let's do it in this way :)

PR #135, opened 2025-08-04
feat: Side panel cy-58

WhoeverWhenever:
- Added mocked wrapper for dashboard/plan page
- Added NavigationItem component to use it in Side Panel and User Menu
- Moving Side Panel items to User Menu when width<1024px
image
image

https://github.com/user-attachments/assets/c029bada-9747-4d97-956b-c6c12f17c925

### DIFF
--- a/apps/frontend/src/pages/dashboard-wrapper-mock/components/wrapper.tsx
+++ b/apps/frontend/src/pages/dashboard-wrapper-mock/components/wrapper.tsx
@@ -1,37 +1,17 @@
-import { type JSX } from "react";
-import { useLocation } from "react-router-dom";
+import { Outlet } from "react-router-dom";
 
 import { AppHeader } from "~/libs/components/components.js";
 import { SidePanel } from "~/libs/components/side-panel/side-panel.js";
-import { AppRoute } from "~/libs/enums/enums.js";
 
 import styles from "../styles.module.css";
-import { Dashboard } from "./dashboard.js";
-import { Plan } from "./plan.js";
 
 const Wrapper: React.FC = () => {
-	const { pathname } = useLocation();
-
-	const getPage = (path: string): JSX.Element => {
-		switch (path) {
-			case AppRoute.DASHBOARD: {
-				return <Dashboard />;
-			}
-
-			case AppRoute.PLAN: {
-				return <Plan />;
-			}
-		}
-
-		return <></>;
-	};
-
 	return (
 		<>
 			<AppHeader />
 			<div className={styles["wrapper"]}>
 				<SidePanel />
-				{getPage(pathname)}
+				<Outlet />
 			</div>
 		</>
 	);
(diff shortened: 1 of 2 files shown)


apps/frontend/src/libs/components/side-panel/side-panel.tsx
```diff
@@ -0,0 +1,73 @@
...
+				}
+			}
+
+			return element as HTMLUListElement;
+		};
+
+		const element = getActiveItem(pathname);
+
+		element.classList.add(styles["active"] as string);
+	}, [childrenOrder.first, childrenOrder.second, pathname]);
```
WhoeverWhenever, line 47: Is there a better way to implement this part?

WhoeverWhenever: I was thinking about onClick() for NavigationItem, but decided to keep this implementation, because in my opinion, it is better to add a class depending on the current path

apps/frontend/src/libs/components/app-header/app-header.tsx
```diff
@@ -69,6 +78,16 @@ const AppHeader: React.FC = () => {
...
 				</div>
+				<button
+					aria-label="Open user menu"
+					className={burgerMenuClassName}
+					onClick={handleMenuToggle}
+					type="button"
+				>
+					<div />
+					<div />
+					<div />
```
WilhelmWesser, line 89: Do you need these?

WilhelmWesser: Please, add at least class names for them so that the purpose of these divs can be understood easily

WilhelmWesser (approved):

roman-mykytka (changes requested):

apps/frontend/src/pages/dashboard-wrapper-mock/components/wrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+	return (
+		<>
+			<AppHeader />
+			<div className={styles["wrapper"]}>
+				<SidePanel />
+				{getPage(pathname)}
+			</div>
+		</>
+	);
+};
```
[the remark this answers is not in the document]
WhoeverWhenever: I took Auth component as an example for a Wrapper. Should we reconsider Auth implementation as well?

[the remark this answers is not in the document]
roman-mykytka: I’d stick to a single approach-routing as the single source of truth.
That means not checking `pathname` in components and not duplicating router logic with local state/switches.
In the case of `Auth`, it was partially reasonable because the component acts as a layout/wrapper with shared logic (navigation init, dispatching actions, ...)

when the local state is appropriate 
a few examples:
- tabs
- accordions  

that don’t need a shareable URL or browser history

roman-mykytka (approved):

apps/frontend/src/pages/dashboard-wrapper-mock/components/wrapper.tsx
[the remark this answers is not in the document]
WhoeverWhenever: Got it. Thank you

PR #166, opened 2025-08-06
fix: remove sign-up white bar cy-164

astowneer:
### On the signup page, a white bar appears at the bottom, and there’s a lack of spacing between the form and the wrapper edges.

Solution:

- Add more spacing around the form and adjust the car image positioning to ensure it doesn't touch or overlap the wrapper edges
- Make orange image bigger to improve the visual effect
- Replace wrong password input filed placeholder to match design pattern

https://github.com/user-attachments/assets/973facd5-2fdb-484e-b7a0-d43f8eff59eb

### DIFF
--- a/apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.module.css
+++ b/apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.module.css
@@ -85,7 +85,7 @@
 		top: 10%;
 		left: -10%;
 		height: 130px;
-		height: clamp(100px, 15vw, 130px)
+		height: clamp(100px, 15vw, 130px);
 	}
 
 	.car-image {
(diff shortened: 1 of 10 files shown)


apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.module.css
```diff
@@ -20,14 +20,19 @@
 	z-index: 200;
 	width: 100%;
```
watchlar2000, line 21: check the class name of this block - it should be .sign-up (not .sign-in)

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -137,7 +137,11 @@ const SignUpForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
 							type="password"
```
watchlar2000, line 137: There's no need to wrap each input with <div className="flow-loose">. Wrap the whole group of inputs inside it instead.

astowneer:
https://github.com/user-attachments/assets/5187f556-e3ec-4ebb-92f6-ed2f891c3d3b

https://github.com/user-attachments/assets/37fb5749-ad94-4e20-8b49-f9e1cf738a09

WilhelmWesser (approved):

Issue #172, opened 2025-08-06
Inconsistent email validation messages between frontend and backend


VictorVagabculov:
## 📝 Description  
When entering an invalid email format in the **Sign In** form, the system may return **two different validation messages** depending on the input:

- If the email contains **double `@` signs** (e.g., `user@@example.com`), the **browser’s built-in HTML5 validation** triggers a default tooltip message.
- If the email contains a malformed but acceptable format for `type="email"` (e.g., `user.@gmail.com`), the **backend returns a custom JSON error**: `"Invalid email format"`.

This results in inconsistent user experience and complicates testing.  
A possible solution is to change the `input type="email"` to `text` and handle validation fully via a custom function.

---

## 🧪 Steps to Reproduce  
1. Navigate to `/sign-in`
2. Enter `user@@domain.com` in the Email field  
3. Enter a valid password, e.g., `ValidPass123`  
4. Click the **Sign In** button  
5. Observe the error message  
6. Repeat steps 2–5 using `user.@domain.com`

---

## 🧾 Test Evidence  

<img width="805" height="769" alt="Image" src="https://github.com/user-attachments/assets/f79c0e23-ba46-40e9-9a7b-b4da9b0b580f" />

<img width="724" height="722" alt="Image" src="https://github.com/user-attachments/assets/d130abdc-17af-4880-b6c3-85412a583eda" />

<img width="360" height="281" alt="Image" src="https://github.com/user-attachments/assets/8bde26e6-bfb1-4f04-a29d-36811b74a8fe" />

<img width="360" height="266" alt="Image" src="https://github.com/user-attachments/assets/ff1680a1-923a-4e2b-904d-1868763e8996" />

PR #173, opened 2025-08-06
fix: inconsistent email validation messages cy-172

muxkuro:
As was proposed in an issue #172 the problem was in the type of email input field. Since we already had error validation on backend, in-built HTML validation overwritten the implemented validation and that caused inconsistent messages on sign-in form. I changed the type of email input field to text, so that we validate the field only through our own implemented validation schema.


VictorVagabculov:
fixed, tested in branch 172-fix-email-validation.

image

PR #196, opened 2025-08-08
fix(frontend): add profile button cy-174

Volin13:
Added profile link at burger menu. Added link on profile image click both headed to /profile - currently is not careated it is a someone`s separated task

chrome_GXPDNa9gpT


daniil-tkalenko:
on your gif this button is present in dropdown menu, but is absent in sidebar on large screen. perhaps it should added there too?

Volin13:
> on your gif this button is present in dropdown menu, but is absent in sidebar on large screen. perhaps it should added there too?
mb) I hesitated if I should add  it on desktop or just add a link on user avatar click. I will fix that

PR #198, opened 2025-08-08
feat: add dashboard screen cy-89

alex-tuz:
Implemented the Dashboard screen, serving as the main interface for authenticated users.

image


apps/frontend/src/libs/components/dashboard-wrapper/dashboard-wrapper.tsx
```diff
@@ -1,20 +1,21 @@
...
 import { SidePanel } from "~/libs/components/side-panel/side-panel.js";
 
-import styles from "../styles.module.css";
+import styles from "./styles.module.css";
 
-const Wrapper: React.FC = () => {
+const DashboardWrapper: FC = () => {
 	return (
-		<>
+		<div className={styles["appContainer"]}>
```
watchlar2000, line 11: Rename appContainer class name to app-container. I think we should follow this class naming convention.

apps/frontend/src/libs/components/dashboard/components/greeting/styles.module.css
```diff
@@ -0,0 +1,88 @@
...
+@media (width <= 1240px) {
+	.button .largeIcon {
+		position: absolute;
+		top: 50%;
+		left: var(--space-l);
+		transform: translateY(-50%) scale(1.5);
+	}
+}
+
+@media (width <= 768px) {
```
watchlar2000, line 68: We should follow mobile first approach, I mean design the UI for small screen first and then use media queries width >= 768px to adapt it to bigger screens.

apps/frontend/src/libs/components/dashboard/components/greeting/styles.module.css
```diff
@@ -0,0 +1,88 @@
...
+
+.button:not(:disabled):hover span[class*="button-icon"] {
+	transform: none;
+}
+
+.button::after {
+	display: none;
+}
+
+@media (width <= 1240px) {
```
watchlar2000, line 59: If you need this particular breakpoint value 1240px, then it would be nice to update breakpoints.css to add this value so other DEVs know that we are using it.

apps/frontend/src/assets/css/fonts.css
```diff
@@ -17,3 +17,13 @@
 			format("truetype-variations");
 	}
 }
+
+@supports (font-variation-settings: normal) {
+	@font-face {
```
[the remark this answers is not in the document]
alex-tuz: @watchlar2000 This font was used in Figma for the button label, so I added it to match the design.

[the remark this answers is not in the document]
watchlar2000: Yeah, you're right, but do we need to add another font just to use it in one place - inside one button? We can suggest improving the design to use existing fonts - for example, we can use some font-weight + vibrant color if that button needs to stand out.

daniil-tkalenko (approved):

apps/frontend/src/libs/components/dashboard/components/create-plan-button/create-plan-button.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: Properties) => {
+	const buttonClasses = getClassNames(styles["button"], className);
+
+	return (
+		<Button
+			className={buttonClasses}
+			disabled={disabled}
+			icon={icon}
+			label={label}
+			{...(onClick ? { onClick } : {})}
```
WilhelmWesser, line 29: We should be able to pass just `onClick` here. Seems that there is no need for this ternary operator

alex-tuz: Made onClick required. Before, it was only used to satisfy exactOptionalPropertyTypes, which didn’t allow passing undefined explicitly.

WilhelmWesser:
@alex-tuz Please, fix conflicts so we can merge your PR

PR #204, opened 2025-08-10
feat: add plan generation loader cy-56

astowneer:
### Progress Tracking Implementation

**Overview**

To accurately track progress, we first need to know the total amount of work involved. Progress is then measured as the portion of completed work relative to this total.

For example, if the plan consists of **4, 7, or 21** days, with **1 to 5** tasks per day, the total number of tasks can be 35 (7 days × 5 tasks). We calculate progress as follows:

progress = (completed_tasks / total_tasks) * 100

**Example:**

* After 1 task: 1 / 35 * 100 ≈ 2.86%
* After 7 tasks: 20%
* After all 35: 100%

After completing all 35 tasks: 100%
_If the plan covers more days, the progress percentage can be calculated more precisely by adjusting the total number of tasks accordingly._

**Options for Implementing Progress Updates:**

**Option A — Fake Progress**

- Backend returns the final plan only after completion.
- Frontend simulates progress by running a timer from 1% up to about 90%, increasing while waiting for the backend response.
- When the backend response arrives, the progress quickly jumps to 100%.

Pros: Very simple to implement; no backend changes needed.
Cons: Does not reflect actual generation progress; purely visual.

**Option B — Store and Poll Progress**

When the frontend sends a request to /plan-generate:

- Backend creates a progress record in a database.
- Backend updates the progress after each completed step (task or day).
- Frontend polls the backend endpoint /progress every second (or at a chosen interval) to retrieve the current progress percentage.

Pros: Reflects real progress; relatively easy to implement.
Cons: Requires managing asynchronous tasks and storage; polling introduces additional requests.

**Option C — Real-Time Updates via WebSockets**

- Similar to Option B, but instead of polling, the backend pushes progress updates to the frontend in real time through WebSockets

Pros: Provides the smoothest and most responsive user experience.
Cons: Requires more complex setup and infrastructure.

**Choosen Approach**

It would be inefficient to track the generation of each individual task. If we don’t track every task, the percentage shown from the backend won’t accurately reflect real progress, which can create a poor user experience. For example, if a plan has 4 days with 5 tasks each (20 tasks total), tracking every single task would result in very small, frequent percentage updates — making the loading feel too jumpy and inconsistent, so for minimal effort and acceptable UX, use Option A with a fake progress bar on the frontend.

https://github.com/user-attachments/assets/c25866a4-d368-4f69-be30-93a71419555c


apps/frontend/src/modules/plans/slices/actions.ts
```diff
@@ -0,0 +1,21 @@
+import { createAsyncThunk } from "@reduxjs/toolkit";
+
+import { type AsyncThunkConfig } from "~/libs/types/types.js";
+import { type PlanDaysTaskDto } from "~/modules/plans/plans.js";
+
+import { type QuizAnswersDto } from "../libs/types/types.js";
+import { name as sliceName } from "./plan.slice.js";
+
+const generate = createAsyncThunk<
```
daniil-tkalenko, line 9:
```suggestion
const generatePlan = createAsyncThunk<
```

apps/frontend/src/pages/plan-generation/plan-generation.tsx
```diff
@@ -0,0 +1,135 @@
...
+			setProgress((previous) => {
+				if (previous >= PROGRESS_MAX_SLOW) {
+					return previous;
+				}
+
+				const distance = PROGRESS_MAX_SLOW - previous;
+				const increment = Math.max(
+					SLOW_INCREMENT_MAX,
+					distance / SLOW_INCREMENT_DIVISOR,
+				);
```
WilhelmWesser, line 77:
```suggestion
				const distance = PROGRESS_MAX_SLOW - previous;
				const increment = Math.max(
					SLOW_INCREMENT_MAX,
					distance / SLOW_INCREMENT_DIVISOR,
				);
```
I would move these calculations to a separate function, something like `getProgressStep`

apps/frontend/src/pages/plan-generation/plan-generation.tsx
```diff
@@ -0,0 +1,135 @@
...
+	useEffect(() => {
+		const stored = localStorage.getItem(StorageKey.QUIZ_ANSWER);
+		const quizAnswers: QuizAnswersDto = stored
+			? (JSON.parse(stored) as QuizAnswersDto)
+			: [];
+
+		void dispatch(planActions.generate(quizAnswers));
+
+		const slowProgressId = setInterval(() => {
+			setProgress((previous) => {
```
WilhelmWesser, line 68: I would also move the whole callback you pass to the setProgress function out of the component and would name it `getUpdatedProgress` or something like that

apps/frontend/src/pages/plan-generation/plan-generation.tsx
```diff
@@ -0,0 +1,135 @@
...
+		};
+	}, [dispatch]);
+
+	useEffect(() => {
+		let fastProgressId: null | ReturnType<typeof setInterval> = null;
+
+		if (status === DataStatus.FULFILLED) {
+			fastProgressId = setInterval(() => {
+				setProgress((previous) =>
+					Math.min(previous + FAST_INCREMENT, PROGRESS_MAX),
```
WilhelmWesser, line 94: This calculation is also worth moving into a separate function because it is not clear what we are calculating in this place specifically

WilhelmWesser (changes requested):

apps/frontend/src/pages/plan-generation/components/slider/libs/constants/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const SLIDE_START_INDEX = 0;
+const SLIDE_INCREMENT = 1;
+
+export {
+	SLIDE_DURATION_MS,
+	SLIDE_INCREMENT,
+	SLIDE_INTERVAL_MS,
+	SLIDE_START_INDEX,
+};
```
roman-mykytka, line 12: I recommend grouping them into logical object-like “enums” based on their usage context to improve readability and avoid confusion

example
separate into slide timing and indexing groups

**B29**

apps/frontend/src/pages/plan-generation/libs/constants/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+const SLOW_INTERVAL_MS = 100;
+const SLOW_INCREMENT_MAX = 0.2;
+const SLOW_INCREMENT_DIVISOR = 20;
+
+const FAST_INTERVAL_MS = 150;
+const FAST_INCREMENT = 1.5;
+
+const PROGRESS_MAX_SLOW = 90;
+const PROGRESS_MIN = 1;
+const PROGRESS_MAX = 100;
```
roman-mykytka, line 22: same **B29**

roman-mykytka (changes requested):

roman-mykytka (approved):

daniil-tkalenko (approved):

WilhelmWesser (approved):

PR #210, opened 2025-08-11
feat: Add 3 more plans styles cy-160

WhoeverWhenever:
- Added PlanStyle as a component
- Added PlanStyle component to the home page
- Weekdays for a "With Remarks" plan style is currently calculated from _created at_ plan date

**Different types of plan styles:**
**With Remarks**

image

**Colourful**

image

**Minimal**

image

**Homepage view**

image

**Choose-style page view**

image

**Dashboard**

https://github.com/user-attachments/assets/89521c99-d8a0-4694-b0b0-1dbda98ffb85


apps/frontend/src/libs/components/plan-styles/mocks/plan-mocks.ts
```diff
@@ -0,0 +1,36 @@
+import { type Plan, type PlanDay, type Task } from "~/libs/types/types.js";
+
+const ZERO = 0;
+const ONE = 1;
```
WilhelmWesser, line 4: As far as I remember we already have ZERO and ONE constants so you can just reuse them

apps/frontend/src/pages/choose-style/choose-style.data.ts
```diff
@@ -1,21 +1,23 @@
...
-	{ id: "box-3", img: ExampleMotivating, label: "Motivating" },
-	{ id: "box-4", img: ExampleColourful, label: "Colourful" },
+	{ id: "box-1", label: "Minimal", planStyle: "minimal" },
+	{
+		id: "box-2",
+		label: "With Remarks",
+		planStyle: "withremarks",
+	},
+	{
+		id: "box-4",
```
WilhelmWesser, line 17: Should not we be sequential? Let's change box-4 to box-3 if we do not use Motivating template anymore

WilhelmWesser (changes requested):

WilhelmWesser (approved):

daniil-tkalenko (approved):

roman-mykytka (approved):

PR #224, opened 2025-08-12
feat: plan generation cy-55

olenkabilonizhka:
- Created OpenAIService responsible for interacting with OpenAI API to generate plans.
- Created system prompt to ensure a structured object response.
- Added GeneratedPlanDTO representing the AI response structure.
- Integrated response validation and regeneration process in case of validation errors.
- Integrated the service into QuizService.

Structure of generated plan:
```
{
    "days": [
        {
            "dayNumber": number, 
            "tasks": [
                {
                    "title": string, 
                    "description": string, 
                    "executionTimeType": "afternoon" | "evening" | "morning" | null, 
                    "order": number
                }
            ]
        }
    ],
    "duration": number,
    "intensity": string,
    "title": string
}
```

### DIFF
--- a/apps/frontend/src/pages/plan-generation/libs/constants/constants.ts
+++ b/apps/frontend/src/pages/plan-generation/libs/constants/constants.ts
@@ -1,13 +1,13 @@
-import {
-	type QuizAnswersRequestDto,
-	type QuizCategoryType,
-} from "../types/types.js";
+import { DataStatus, QuizIndexes } from "~/libs/enums/enums.js";
+import { type QuizState } from "~/modules/quiz/slices/quiz.slice.js";
 
-const DEFAULT_CATEGORY: QuizCategoryType = "hobby";
-const DEFAULT_QUIZ_ANSWERS_PAYLOAD: QuizAnswersRequestDto = {
-	answers: [],
-	category: DEFAULT_CATEGORY,
+const DEFAULT_QUIZ_STATE: QuizState = {
+	answers: {},
+	currentQuestion: QuizIndexes.FIRST_INDEX,
+	dataStatus: DataStatus.IDLE,
 	notes: "",
+	questions: null,
+	selectedCategory: null,
 };
 
 const SlowTiming = {
@@ -30,7 +30,7 @@ const ProgressLimits = {
 const LOADING_DURATION_MS = 6000;
 
 export {
-	DEFAULT_QUIZ_ANSWERS_PAYLOAD,
+	DEFAULT_QUIZ_STATE,
 	FastTiming,
 	LOADING_DURATION_MS,
 	ProgressLimits,
--- a/apps/frontend/src/pages/plan-generation/plan-generation.tsx
+++ b/apps/frontend/src/pages/plan-generation/plan-generation.tsx
@@ -1,6 +1,5 @@
 import { useEffect } from "react";
 import { useNavigate } from "react-router-dom";
-import { type QuizAnswersRequestDto } from "shared";
 
 import {
 	Book,
@@ -19,10 +18,15 @@ import { getClassNames } from "~/libs/helpers/get-class-names.js";
 import { useAppDispatch, useAppSelector } from "~/libs/hooks/hooks.js";
 import { storage, StorageKey } from "~/libs/modules/storage/storage.js";
 import { actions as planActions } from "~/modules/plans/plans.js";
+import { actions as quizActions } from "~/modules/quiz/quiz.js";
 import { type QuizState } from "~/modules/quiz/slices/quiz.slice.js";
+import {
+	type QuizAnswersRequestDto,
+	type QuizCategoryType,
+} from "~/pages/plan-generation/libs/types/types.js";
 
 import { ImageSlider } from "./components/slider/slider.js";
-import { DEFAULT_QUIZ_ANSWERS_PAYLOAD } from "./libs/constants/constants.js";
+import { DEFAULT_QUIZ_STATE } from "./libs/constants/constants.js";
 import { useProgress } from "./libs/hooks/hooks.js";
 import styles from "./styles.module.css";
 
@@ -48,18 +52,15 @@ const PlanGeneration: React.FC = () => {
 	useEffect(() => {
 		const generatePlan = async (): Promise<void> => {
 			const stored = await storage.get(StorageKey.QUIZ_STATE);
-			const storedObject = stored
-				? (JSON.parse(stored) as Omit<QuizState, "dataStatus">)
-				: undefined;
+			const quizState = stored
+				? (JSON.parse(stored) as QuizState)
+				: DEFAULT_QUIZ_STATE;
 
-			const quizAnswers: QuizAnswersRequestDto =
-				storedObject && storedObject.selectedCategory
-					? {
-							answers: Object.values(storedObject.answers),
-							category: storedObject.selectedCategory,
-							notes: storedObject.notes,
-						}
-					: DEFAULT_QUIZ_ANSWERS_PAYLOAD;
+			const quizAnswers: QuizAnswersRequestDto = {
+				answers: Object.values(quizState.answers),
+				category: quizState.selectedCategory as QuizCategoryType,
+				notes: quizState.notes,
+			};
 
 			await dispatch(planActions.generatePlan(quizAnswers));
 		};
@@ -68,7 +69,11 @@ const PlanGeneration: React.FC = () => {
 	}, [dispatch]);
 
 	const progress = useProgress({
-		onComplete: () => void navigate(AppRoute.PLAN),
+		onComplete: (): void => {
+			dispatch(quizActions.resetQuiz());
+			void storage.drop(StorageKey.QUIZ_STATE);
+			void navigate(AppRoute.PLAN);
+		},
 		status,
 	});
 


apps/backend/src/modules/openai/libs/utilities/assistant-prompt.ts
```diff
@@ -0,0 +1,8 @@
+const generateAssistantPrompt = (
+	errorMessage: string,
+	previousResponse: string,
+): string => {
+	return `The previous response contains the following error: ${errorMessage}. Please fix it.\nPrevious response:\n${previousResponse}`;
```
muxkuro, line 5: I see that we return just formated string for prompts. Maybe it's better to create seperate class like PromptBuilder? With methods like addContext, addError and in the end it will look like this?

```
const generateAssistantPrompt = (
	errorMessage: string,
	previousResponse: string,
) => PromptBuilder.create().addError('Instruction', `Response container error: ${errorMessage}`)
export { generateAssistantPrompt };
```

Also you have system-prompt.ts Maybe consider moving this return message to a seperate enum?

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+		});
+		this.maxAttempts = MAX_ATTEMPTS;
+	}
+
+	public async generatePlan(
+		userPrompt: string,
+		attempts: number = ZERO,
+		assistantPrompt?: string,
+	): Promise<GeneratedPlanDTO> {
+		let answer = "";
```
muxkuro, line 38: Do we need ?
```
let answer = "" 
```

Why not declare answer as const in try catch block?

```
try{
const answer = await this.callAPI
```

I don't see it getting changed anywhere in the current scope, so it can be turned into const?

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+			});
+
+			const planData = JSON.parse(answer) as GeneratedPlanDTO;
+
+			this.validatePlan(planData);
+
+			return planData;
+		} catch (error) {
+			if (attempts + ONE < this.maxAttempts) {
+				// Delay before next retry
```
muxkuro, line 54: No comments in the code. Code should explain itself.

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+			messages: [
+				{ content: userPrompt, role: OpenAIRoles.USER },
+				{ content: systemPrompt, role: OpenAIRoles.SYSTEM },
+				...(assistantPrompt
+					? [{ content: assistantPrompt, role: OpenAIRoles.ASSISTANT }]
+					: []),
+			],
+			model: config.ENV.OPEN_AI.TEXT_GENERATION_MODEL,
+			response_format: { type: "json_object" },
+			temperature: 0.6,
```
muxkuro, line 92: Magic number. Consider moving it to a seperate constant?

apps/backend/src/modules/openai/libs/utilities/assistant-prompt.ts
olenkabilonizhka: Thanks for the PromptBuilder idea, I’ll think how to organize it.
Do you mean creating an enum with a single value for systemPrompt? Why we need to use enum, but not const in this case?

apps/backend/src/modules/openai/openai.service.ts
WilhelmWesser: Well, the logic seems to be the following:
1. We make a request to an LLM model. If it succeeds, we assign the answer variable a value, process the answer and then return the planData.
2. In case the request to the model fails we call the `generatePlan()` function inside the catch block with a delay. In case the the answer variable will be empty.
3. In case the generated plan validation fails we take the model output (answer) and ask the model to fix it with another call to `generatePlan()`

Do I miss something?

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+				return await this.generatePlan(
+					userPrompt,
+					attempts + ONE,
+					generateAssistantPrompt(errorMessage, answer),
+				);
+			}
+
+			throw new Error(PlanErrorMessages.GENERATION_FAILED);
+		}
```
WilhelmWesser, line 70:
```suggestion
			if (attempts + ONE >= this.maxAttempts) {
             throw new Error(PlanErrorMessages.GENERATION_FAILED);
			}
     // Here is the code that you put inside if clause previously
```
Let's change this condition block into early quit.

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+		const response = await this.openai.chat.completions.create({
+			messages: [
+				{ content: userPrompt, role: OpenAIRoles.USER },
+				{ content: systemPrompt, role: OpenAIRoles.SYSTEM },
+				...(assistantPrompt
+					? [{ content: assistantPrompt, role: OpenAIRoles.ASSISTANT }]
+					: []),
+			],
+			model: config.ENV.OPEN_AI.TEXT_GENERATION_MODEL,
+			response_format: { type: "json_object" },
```
WilhelmWesser, line 91: Let's move this response type to an enum

WilhelmWesser (changes requested):

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,141 @@
...
+import {
+	generateAssistantPrompt,
+	systemPrompt,
+} from "./libs/utilities/utilities.js";
+import {
+	planCreateValidationSchema,
+	taskCreateValidationSchema,
+} from "./libs/validation-schemas/validation-schemas.js";
+
+class OpenAIService {
```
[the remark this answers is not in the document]
olenkabilonizhka: It's main openai service, so it makes sense to keep it in apps/backend/src/modules/openai/openai.service.ts. What do you think?

apps/backend/src/modules/openai/openai.service.ts
olenkabilonizhka: > Well, the logic seems to be the following:
> 
> 1. We make a request to an LLM model. If it succeeds, we assign the answer variable a value, process the answer and then return the planData.
> 2. In case the request to the model fails we call the `generatePlan()` function inside the catch block with a delay. In case the the answer variable will be empty.
> 3. In case the generated plan validation fails we take the model output (answer) and ask the model to fix it with another call to `generatePlan()`
> 
> Do I miss something?

Yes, exactly. Since the last changes, I added sending errors to the frontend when we receive certain errors (400, 401, 404) from the model.

apps/backend/src/modules/openai/libs/constants/schemas.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line quotes
+const TASK_SCHEMA = `{"title": string, "description": string, "executionTimeType": "afternoon" | "evening" | "morning" | null, "order": number}`;
+
```
daniil-tkalenko, line 3:
```suggestion
// eslint-disable-next-line quotes
const TASK_SCHEMA = "{\"title\": string, \"description\": string, \"executionTimeType\": \"afternoon\" | \"evening\" | \"morning\" | null, \"order\": number}";

```

apps/backend/src/modules/openai/libs/enums/openai-roles.ts
```diff
@@ -0,0 +1,7 @@
+const OpenAIRoles = {
+	ASSISTANT: "assistant",
+	SYSTEM: "system",
+	USER: "user",
+} as const;
+
```
daniil-tkalenko, line 6:
```suggestion
const OpenAIRole = {
	ASSISTANT: "assistant",
	SYSTEM: "system",
	USER: "user",
} as const;

```

https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md
B23

please update other places too

apps/backend/src/modules/openai/libs/enums/plan-error-messages.ts
```diff
@@ -0,0 +1,12 @@
+const PlanErrorMessages = {
```
daniil-tkalenko, line 1: like here

apps/backend/src/modules/openai/libs/utilities/prompt-builder.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	addPreviousResponse(response: string): this {
+		this.prompt.push(`Previous response:\n${response}`);
+
+		return this;
+	}
+
+	build(): string {
+		return this.prompt.join("\n");
+	}
```
daniil-tkalenko, line 30: https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md
B30
please update all other places as well

daniil-tkalenko (changes requested):

packages/shared/src/modules/openai/libs/types/generated-plan-dto.type.ts
```diff
@@ -0,0 +1,24 @@
+import { type ValueOf } from "src/libs/types/types.js";
+
+import { type ExecutionTimeType } from "../../../tasks/tasks.js";
+
+type GeneratedDayDTO = {
```
astowneer, line 5: We already have how response should look like `shared/src/modules/plans/libs/types/plan-dto.type.ts`, so I think this types is not needed

apps/backend/src/modules/quiz/quiz.service.ts
```diff
@@ -1,53 +1,26 @@
...
-	public handleAnswers(payload: QuizAnswersRequestDto): unknown {
+	private openAIService: OpenAIService;
+
+	public constructor() {
+		this.openAIService = openAiService;
+	}
+
+	public async handleAnswers(
+		payload: QuizAnswersRequestDto,
+	): Promise<GeneratedPlanDTO> {
```
astowneer, line 18: Return `PlanDaysTaskDto` from` packages/shared/src/modules/plans/libs/types/plan-dto.type.ts`

apps/backend/src/modules/openai/openai.service.ts
```diff
@@ -0,0 +1,154 @@
...
+			);
+		}
+	}
+
+	private async callAPI({
+		assistantPrompt,
+		systemPrompt,
+		userPrompt,
+	}: {
+		assistantPrompt: string | undefined;
```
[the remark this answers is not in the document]
olenkabilonizhka: "exactOptionalPropertyTypes": true doesn’t allow me to use this

apps/backend/src/modules/quiz/quiz.service.ts
olenkabilonizhka: We can’t use it because of ids, which are required. There’s no user id if the user isn’t authenticated and the other ids will be generated in the db, isn’t it? I don’t think mocking ids would be the right approach.

packages/shared/src/modules/openai/libs/types/generated-plan-dto.type.ts
olenkabilonizhka: Answered in comment regarding PlanDaysTaskDto

apps/backend/src/modules/openai/libs/constants/schemas.ts
```diff
@@ -0,0 +1,8 @@
...
+const DAY_SCHEMA = `{"dayNumber": number,"tasks": [${TASK_SCHEMA}]}`;
+
+const PLAN_SCHEMA = `{"days": [${DAY_SCHEMA}],"duration": number,"intensity": string,"title": string}`;
+
```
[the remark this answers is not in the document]
olenkabilonizhka: I think it’s unnecessary to write a constant and then map it into a string if that constant isn’t used anywhere else.

daniil-tkalenko (approved):

PR #243, opened 2025-08-13
feat: add profile page cy-176

alex-tuz:


apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,171 @@
...
+			<div className={styles["profile-container"]}>
+				<header className="flow">
+					<h1 className={getClassNames(styles["title"])} id="profile-title">
+						Profile
+					</h1>
+				</header>
+				<form
+					aria-labelledby="profile-title"
+					className={getClassNames(sharedStyles["form"], "cluster")}
+					// eslint-disable-next-line react/jsx-no-bind
```
daniil-tkalenko, line 107: find a way to implement this without commenting roles, consider extracting logic to handler function

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,171 @@
...
+	const { user } = useAppSelector(({ auth }) => auth);
+
+	const defaultValues = useMemo<UserUpdateRequestDto>(
+		() => ({
+			confirmPassword: "",
+			dob: formatDateForInput(user?.dob ?? null),
+			email: user?.email ?? "",
+			name: user?.name ?? "",
+			password: "",
+		}),
```
daniil-tkalenko, line 66: extract to getDefaultValues function

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,171 @@
...
+	}, [defaultValues, reset]);
+
+	const handleFormSubmit = handleSubmit((data) => {
+		const changed = getChangedProfileFields(data, defaultValues);
+
+		if (Object.keys(changed).length === ZERO) {
+			return;
+		}
+
+		void dispatch(updateProfile(changed));
```
daniil-tkalenko, line 87: i would suggest to remove this logic, and instead send whole form values to BE on form submit. but it would be nice to disable form from submitting when there were no changes to it; you could use isDirty prop from useAppForm for that

packages/shared/src/modules/users/libs/types/user-update-request-dto.type.ts
```diff
@@ -1,7 +1,9 @@
 type UserUpdateRequestDto = {
-	email?: string;
-	name?: string;
-	password?: string;
+	confirmPassword?: string | undefined;
+	dob?: string | undefined;
+	email?: string | undefined;
+	name?: string | undefined;
+	password?: string | undefined;
```
daniil-tkalenko, line 6: create this type from zod schema, instead of manual creation

packages/shared/src/modules/users/libs/validation-schemas/user-update.validation-schema.ts
```diff
@@ -0,0 +1,65 @@
...
+		email: z
+			.string()
+			.trim()
+			.min(UserValidationRule.NON_EMPTY_STRING_MIN_LENGTH, {
+				message: UserValidationMessage.FIELD_REQUIRED,
+			})
+			.regex(UserValidationRegexRule.EMAIL_VALID_CHARS_MIN_MAX, {
+				message: UserValidationMessage.EMAIL_INVALID,
+			})
+			.optional(),
```
daniil-tkalenko, line 29: i not sure email can be optional 
I think it required for every user

packages/shared/src/modules/users/libs/validation-schemas/user-update.validation-schema.ts
```diff
@@ -0,0 +1,65 @@
...
+			.refine(
+				(value) => UserValidationRegexRule.NAME_VALID_SURROUNDING.test(value),
+				{
+					message: UserValidationMessage.NAME_SURROUNDING_RULE,
+				},
+			)
+			.refine((value) => UserValidationRegexRule.NAME_VALID_CHARS.test(value), {
+				message: UserValidationMessage.NAME_ONLY_ALLOWED_CHARS,
+			})
+			.optional(),
```
daniil-tkalenko, line 50: and I think name as well

apps/backend/src/modules/users/user.service.ts
```diff
@@ -60,7 +65,35 @@ class UserService implements Service {
 		id: number,
 		payload: UserUpdateRequestDto,
 	): Promise<null | UserDto> {
-		const item = await this.userRepository.update(id, payload);
+		if (payload.email) {
```
daniil-tkalenko, line 68: once other comments will be fixed, fix this logic as well, since email will be always in payload.

apps/backend/src/modules/users/user.service.ts
```diff
@@ -60,7 +65,35 @@ class UserService implements Service {
...
+			}
+		}
+
+		const updateData = Object.fromEntries(
+			Object.entries({
+				dob: payload.dob,
+				email: payload.email,
+				name: payload.name,
+			}).filter(([, value]) => value !== undefined),
+		) as Partial<UserModel>;
```
daniil-tkalenko, line 88: inspect this part as well, and adjust if in the end some parts will look a bit redundant

daniil-tkalenko (changes requested):

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -38,6 +41,20 @@ class UserController extends BaseController {
 			method: HTTPRequestMethod.GET,
 			path: UsersApiPath.ROOT,
 		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.update(
+					options as APIBodyOptions<UserUpdateRequestDto> & {
+						user?: { id: number };
```
WilhelmWesser, line 49: As far as I remember we already have a type for `{ id: number }` in the app

WilhelmWesser: Check it and if we already have such type replace every `{ id: number }` in your PR

apps/frontend/src/libs/helpers/date-helpers.ts
```diff
@@ -0,0 +1,11 @@
+const DATE_PART_INDEX = 0;
+
+const formatDateForInput = (dateString: null | string): string => {
+	if (!dateString) {
+		return "";
+	}
+
+	return new Date(dateString).toISOString().split("T")[DATE_PART_INDEX] ?? "";
```
WilhelmWesser, line 8: to make it look better I would suggest to change the destructuring approach to
`
const [date] = new Date(dateString).toISOString().split("T")
return date ?? ""
`

apps/frontend/src/modules/users/users-api.ts
```diff
@@ -29,6 +33,20 @@ class UserApi extends BaseHTTPApi {
...
 		return await response.json<UserGetAllResponseDto>();
 	}
+
+	public async updateMe(payload: UserUpdateRequestDto): Promise<UserDto> {
+		const response = await this.load(
+			this.getFullEndpoint(UsersApiPath.ME, {}),
+			{
+				contentType: ContentType.JSON,
+				hasAuth: true,
+				method: "POST",
```
WilhelmWesser, line 43: Use and enum for REST Method name

WilhelmWesser (changes requested):

WilhelmWesser (approved):

PR #275, opened 2025-08-17
feat: add task editing/deletion and day deletion

muxkuro:


apps/frontend/src/modules/tasks/mocks/handlers/task.handler.ts
```diff
@@ -0,0 +1,59 @@
...
+		const taskIndex = MOCK_TASKS.findIndex((task) => task.id === taskId);
+
+		if (taskIndex === TaskConstants.TASK_INDEX_NOT_FOUND) {
+			return new Response(
+				JSON.stringify({ message: TaskMessage.TASK_NOT_FOUND }),
+				{ status: HTTPCode.NOT_FOUND },
+			);
+		}
+
+		MOCK_TASKS[taskIndex] = {
```
WilhelmWesser, line 31: You should not mutate imported entities even if they are mocks

WilhelmWesser (changes requested):

apps/frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,49 @@
...
+	isOpen,
+	onClose,
+	title,
+}: Properties) => {
+	if (!isOpen) {
+		return null;
+	}
+
+	return (
+		<div aria-modal="true" className={styles["modal-overlay"]} role="dialog">
```
watchlar2000, line 28: https://github.com/BinaryStudioAcademy/bsa-2025-checkly/pull/283/files#r2285283561

There's another PR adding the same component.

Instead of div, consider using Dialog native element.

WilhelmWesser (approved):

PR #285, opened 2025-08-19
refactor(frontend): improve logout redirect logic cy-169

JosephAparicio:
Problem:

"I see that you changed the route of the redirect to AppRoute.ROOT
But it only works if the user clicks Logout on a public route, otherwise it redirects the user to sign-in.
This happens because of the protected routes, as soon as the user is unauthenticated the component rerenders and identifies that, redirecting the user before the actual redirect you implemented triggers"

Solutions proposed by the QA team:

- Create a public /logout route where the user gets redirected, and that page logs you out and then redirects you to Home (very quickly).
- Create a global logging_out state, which gets activated by the logout action, and in protected routes add an if logging_out is active → Navigate to="/signin" replace.

New flow:

- User clicks "Logout" from any route (public or protected).
- The app navigates to a dedicated /logout page.
- The /logout page dispatches the logout action to clear authentication state.
- After logout, the user is always redirected to the Home page, regardless of the previous route.
- Protected routes no longer interfere with the intended redirect after logout.
- The sign-in page loader only appears during actual authentication requests (not on initial load).

https://github.com/user-attachments/assets/3e7e7c8b-c4b8-4388-aa81-1b0639323b27


package-lock.json
```diff
@@ -50,7 +50,7 @@
 			}
 		},
 		"apps/backend": {
-			"version": "1.5.0",
```
WilhelmWesser, line 53: If you do not install any new dependencies package-lock,json file must contain no changes

WilhelmWesser (changes requested):

Issue #269, opened 2025-08-16
Refactor GET /quiz-questions to return category-specific options for the first question


brunoazvd:
## Description

Currently, the endpoint `GET /quiz-questions` returns a static array of questions, regardless of the category selected at the beginning of the quiz. Specifically we need to enhance the first question options: "What is your current area of focus?

For example, when a user selects the "Sport" category, the options include unrelated items such as "🎨 Creativity & Art", which do not fit within the chosen category.

As highlighted by the PO during last demo, this breaks the logic of the quiz and reduces the sense of personalization for the user.

**The first question should still be same for every category: "What is your current area of focus?"**. But the options should depend on the category itself.

## Expected Behavior

- The first quiz question should adapt dynamically based on the category selected by the user.
- Options must be relevant to the chosen category.
- Endpoint should support receiving the category to return the appropriate set of questions.

## Acceptance Criteria

- `GET /quiz-questions` accepts a category identifier.
- The first question’s options should be relevant to the chosen category and irrelevant options are no longer shown.

PR #313, opened 2025-08-20
feat: category specific questions cy-269

watchlar2000:
Hello!

Ref ticket #269 

## Below is the list of changes:

**Backend:**

- Enhanced `/quiz-questions` endpoint to accept a `categoryId` query parameter to fetch category-specific questions.

- Questions seed has been extended to add category-specific data.

**Frontend**

- Updated Category page to fetch categories data and render it on the page. Also, when user selects the category and moves to the quiz, quiz-start page receives all the category-related questions.

## Demo

https://github.com/user-attachments/assets/514763d5-09bc-403b-b906-93f49f2e1fc5

### DIFF
--- a/apps/frontend/src/modules/quiz/slices/quiz.slice.ts
+++ b/apps/frontend/src/modules/quiz/slices/quiz.slice.ts
@@ -63,7 +63,7 @@ const { actions, name, reducer } = createSlice({
 		},
 		saveAnswer: (state, action: { payload: QuizAnswer }) => {
 			const answer = action.payload;
-			state.answers[answer.questionId] = answer;
+			state.answers[state.currentQuestion] = answer;
 		},
 		setCategory: (state, action: { payload: string }) => {
 			state.selectedCategory = action.payload;


packages/shared/src/modules/quiz/libs/validation-schemas/question-category-search-parameters.validation-schema.ts
```diff
@@ -0,0 +1,30 @@
+import z from "zod";
+
+import { QuestionValidationMessage } from "../enums/enums.js";
+
+const ZERO = 0;
```
roman-mykytka, line 5: I believe we already have a shared constant for 0 — I’ve seen it in other PRs quite often. It would be better to reuse the existing one from shared, or move this into shared if it doesn’t exist yet.

roman-mykytka (approved):

package-lock.json
```diff
@@ -54,9 +54,8 @@
 			}
```
WilhelmWesser, line 54: As you did no updates to package.json file there should be no changes in the lock file

WilhelmWesser (changes requested):

package-lock.json
watchlar2000: > As you did no updates to package.json file there should be no changes in the lock file

I haven't installed any dependencies. Is it possible this updated package-lock.json comes from another commit? I've rebased a few times to add updates from the main branch.

PR #316, opened 2025-08-21
feat(frontend): add choose styles flow cy-299

muxkuro:
1. Created a seperate enum with current available styles
2. Only authenticated user can access choose-style page
3. Added a reducer to the current plan slice so we can share the current style between pages
4. Removed previous buttons, so that only "Save" button and "Back" arrow button are left
5. After choosing the style and clicking on the "Save" button the user is redirected to the Preview page with set style saved
6. Back button takes user back to Preview page and discards changes

save-style

### DIFF
--- a/apps/frontend/src/pages/choose-style/choose-style.tsx
+++ b/apps/frontend/src/pages/choose-style/choose-style.tsx
@@ -69,19 +69,21 @@ const ChooseStyle: React.FC = () => {
 			>
 				<div className={styles["nav"]}>
 					<NavLink className={navLink} to={AppRoute.OVERVIEW_PAGE}>
-						<button aria-label="Go back">
+						<button aria-label="Go back" className={styles["nav-back-button"]}>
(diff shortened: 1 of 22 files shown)


apps/frontend/src/libs/enums/plan-style-options.enum.ts
```diff
@@ -0,0 +1,7 @@
+const PlanStyleOptions = {
+	COLOURFUL: "COLOURFUL",
+	MINIMAL: "MINIMAL",
+	WITH_REMARKS: "WITH_REMARKS",
+} as const;
+
+export { PlanStyleOptions };
```
daniil-tkalenko, line 7:
```suggestion
const PlanStyle = {
	COLOURFUL: "COLOURFUL",
	MINIMAL: "MINIMAL",
	WITH_REMARKS: "WITH_REMARKS",
} as const;

export { PlanStyle };
```

daniil-tkalenko (approved):

PR #320, opened 2025-08-21
feat: saving generated plan to db automatically cy-309

olenkabilonizhka:
Implemented automatic saving of generated plan to the database:
- Generated plans are now saved automatically.
- If the user is unauthenticated, the plan is also saved with `userId = null`.
- Generated plan id is stored in localStorage for unauthenticated users.

Other updates:
- Modified plan table to allow `userId` to be null.
- Added `key` field to the plan_categories table.
- Updated task edit and delete operations to interact correctly with the database.
- Improved authentication logic to retrieve `userId` even on public pages.


apps/backend/src/plugins/authorization/authorization.plugin.ts
```diff
@@ -30,23 +30,26 @@ const authStrategy = "Bearer ";
...
 			});
 		}
 
-		const tokenValue = authorization.replace(authStrategy, "");
+		const tokenValue = authorization?.replace(authStrategy, "");
 
-		const payload = (await token.decode(tokenValue)) as { userId: number };
+		const payload = tokenValue
+			? ((await token.decode(tokenValue)) as { userId: number })
+			: { userId: null };
```
daniil-tkalenko, line 48: instead of add these changes to auth plugin, why can't we just send userId from auth slice to generate Plan endpoint via payload, if user is authed?

apps/backend/src/db/migrations/20250727185929_plans_days_task_tables.ts
```diff
@@ -65,7 +65,6 @@ async function up(knex: Knex): Promise<void> {
 		table.string(PlanColumnName.TITLE, ColumnLength.TITLE).notNullable();
 		table
 			.integer(PlanColumnName.USER_ID)
-			.notNullable()
```
WilhelmWesser, line 68: We should not modify old migrations to avoid inconsistency. Do it in a new migration

apps/backend/src/db/migrations/20250814182054_add_plan_categories_table.ts
```diff
@@ -30,6 +31,7 @@ async function up(knex: Knex): Promise<void> {
 	await knex.schema.createTable(TABLE_NAMES.PLAN_CATEGORIES, (table) => {
 		table.increments(CategoriesColumnName.ID).primary();
 		table.string(CategoriesColumnName.TITLE).notNullable();
+		table.string(CategoriesColumnName.KEY).notNullable();
```
WilhelmWesser, line 34: Same here

WilhelmWesser (changes requested):

WilhelmWesser (approved):

PR #330, opened 2025-08-22
feat: Forgot Password cy-227

WhoeverWhenever:
[RESEND ](https://resend.com/api-keys) is used as an email service. It is easy to use, but without a custom domain, you can send emails only to your email address

To complete the task, I used some ideas from these articles:
https://supertokens.com/blog/implementing-a-forgot-password-flow
https://security.stackexchange.com/questions/213975/how-to-properly-create-a-password-reset-token
https://security.stackexchange.com/questions/199250/storing-password-reset-token-in-db-what-are-the-implications

Some features:
- Successful toast about sending an email fires on both fulfilled and rejected states (to prevent attackers from knowing some emails are invalid);
- If the token is invalid, the user is redirected to the homepage
- Only the last created token is valid

Possible enhancements:
- Run a job to delete invalid tokens in the db
- Use React Template for an email message
- Use a custom domain to send emails to different addresses

https://github.com/user-attachments/assets/105c4bc5-da9f-4da8-81e0-d872fe5a4d6e


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,24 +1,88 @@
...
+	private emailService: EmailService;
+
 	private encryptor: Encryptor;
+
+	private passwordTokenService: PasswordTokenService;
+
 	private userService: UserService;
 
-	public constructor(userService: UserService, encryptor: Encryptor) {
+	// eslint-disable-next-line max-params
```
WilhelmWesser, line 33: I think you can easily avoid the error if you pass parameters to the constructor as a "bag of properties"
I mean `{
		userService: UserService,
		encryptor: Encryptor,
		emailService: EmailService,
		passwordTokenService: PasswordTokenService,
}`

package-lock.json
```diff
@@ -54,9 +54,8 @@
 			}
 		},
 		"apps/backend": {
-			"version": "1.8.1",
+			"version": "1.9.0",
```
WilhelmWesser, line 57: Take version back from main

package-lock.json
```diff
@@ -87,7 +87,7 @@
 			}
 		},
 		"apps/frontend": {
-			"version": "1.17.1",
+			"version": "1.19.1",
```
WilhelmWesser, line 90: Take version back from main

package-lock.json
```diff
@@ -14982,7 +14979,7 @@
 			}
 		},
 		"packages/shared": {
-			"version": "1.9.0",
+			"version": "1.10.0",
```
WilhelmWesser, line 14982: Take version back from main

WilhelmWesser (changes requested):

PR #350, opened 2025-08-25
fix: repeated styles in sample layouts cy-342

Amicuchi:
**What was done?**

The constant layoutExamples, which serves as the data source for the section, was updated to contain only unique styles, eliminating the duplicate entries.

**Why was it done?**

This change addresses the issue described in the task 342, where the same three styles were repeated throughout the list, creating a poor visual impression.

For this update, I just removed the duplicate styles.
Removing the slider didn’t seem ideal, since it’s functional and can still be used to accommodate new styles in the future.

### DIFF
--- a/apps/frontend/src/pages/home/components/visual-layouts/libs/constants.ts
+++ b/apps/frontend/src/pages/home/components/visual-layouts/libs/constants.ts
@@ -18,26 +18,6 @@ const layoutExamples: LayoutExample[] = [
 		planStyle: "MINIMAL",
 		title: "Minimal",
 	},
-	{
-		id: 4,
-		planStyle: "WITH_REMARKS",
-		title: "With Remarks",
-	},
-	{
-		id: 5,
-		planStyle: "COLOURFUL",
-		title: "Colourful",
-	},
-	{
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/home/components/visual-layouts/libs/constants.ts
```diff
@@ -18,26 +18,6 @@ const layoutExamples: LayoutExample[] = [
...
-	{
-		id: 4,
-		planStyle: "WITH_REMARKS",
-		title: "With Remarks",
-	},
-	{
-		id: 5,
-		planStyle: "COLOURFUL",
-		title: "Colourful",
-	},
```
daniil-tkalenko, line 30: It seems that you removed colorful theme too, not only duplicates, is it intended?

Amicuchi: nope.
The colourful it's in line 15.
This one specifically uses only one line. Seems to be hide. :-)

daniil-tkalenko (approved):

Issue #351, opened 2025-08-25
fix: blank space at the bottom of Dashboard page when there's no active plan


spadaivan:
## Description
Dashboard page has white background instead of grid pattern below plans section if the user doesn't have any plan

## Steps to Reproduce
1. Go to [Checkly](http://checkly.eu-north-1.elasticbeanstalk.com/).
2. Login or Register.
3. Go to [Dashboard](http://checkly.eu-north-1.elasticbeanstalk.com/dashboard).
4. Make sure you have no active plans. If you have, create a new account.
5. Observe the bottom of the page.

## Environment

- Device: Desktop
- OS: Windows 11
- Browser: Chrome 139.0.7258.139
- Environment: Production
- Date: 2025-08-25

## Actual Result
Blank space left at the bottom of the page

## Expected Result
The entire background should have the same grid pattern

## Test Evidence

<img width="1817" height="1257" alt="Image" src="https://github.com/user-attachments/assets/049a8fcd-84d7-431d-8982-4ccb470bbbe5" />

brunoazvd:
I will add a few requests to this ticket, instead of creating a new one, to avoid regression errors later.

- Remove "Create Plan" and "No active plan found" in the **Current active plan** section.
- Instead of **"Current active plan"** it should say **"You have no plans yet."** when the user has no plans.
- Enable to **"Create New Plan"** button that is present to the right of the welcome message. It should redirect the user to `/quiz` and be enabled regardless of plan amount.

Here's a quick mockup to help you understand:
<img width="1639" height="909" alt="Image" src="https://github.com/user-attachments/assets/28865bec-7dae-4f01-afca-c18a74d7aba9" />

PR #361, opened 2025-08-25
fix(frontend): fix blank space, enable button and show new message cy-351

JosephAparicio:
Before Result and Suggestions:

- Blank space left at the bottom of the page
- Remove "Create Plan" and "No active plan found" in the Current active plan section.
- Instead of "Current active plan" it should say "You have no plans yet." when the user has no plans.
- Enable to "Create New Plan" button that is present to the right of the welcome message. It should redirect the user to /quiz and be enabled regardless of plan amount.

Actual Result:
Every point is visible in this video

https://github.com/user-attachments/assets/9b685f4a-3d2b-4304-ae28-c33107d377e6


brunoazvd:
I've tested the changes on the branch and everything that we've asked in the #351 is now fixed.

WilhelmWesser (approved):

PR #374, opened 2025-08-26
fix: password reset endpoint requiring current password + duplicate success toasts cy-371

olenkabilonizhka:
- Removed duplicate code in listenerMiddleware.
- Removed validation for current password — now it updates the password correctly.
- Updated the from address for sent emails.

image

image

image


brunoazvd (commented):
Both issues are fixed and sender email domain modified correctly, I'm able to receive reset password emails now

daniil-tkalenko:
@olenkabilonizhka @brunoazvd it looks like there were some changes to the logic added previusly to handle update user details and change password from profile on dashboard. won't this changes broke this functionality in any way?

brunoazvd:
Retested after last commit and everything is working:
I was able to reset password, I'm also able to update every field in the profile without issues: password, name, email, dob

PR #386, opened 2025-08-27
fix: plan regeneration cy-382

astowneer:


apps/frontend/src/modules/plans/plans-api.ts
```diff
@@ -45,7 +45,7 @@ class PlanApi extends BaseHTTPApi {
 			this.getFullEndpoint(PlansApiPath.GENERATE, {}),
 			{
 				contentType: ContentType.JSON,
-				hasAuth: false,
+				hasAuth: true,
```
watchlar2000, line 48: shouldn't plan generation be public?

daniil-tkalenko (approved):

PR #413, opened 2025-08-28
feat(frontend/backend/shared): add pdf export cy-348

muxkuro:
1. Implemented PDF export with real plan data.
2. Fixed Desktop PNG export to use real plan data.
3. Fixed plan-style-overview, plan-style-print, current-plan, past-plans components to use real plan data

## Plan style overview
image

## Exported PDF plan
[21-Day Spiritual Growth Journey.pdf](https://github.com/user-attachments/files/22019466/21-Day.Spiritual.Growth.Journey.pdf)

## Exported PNG plan divided into 3 parts
21-Day Spiritual Growth Journey-part-3
21-Day Spiritual Growth Journey-part-2
21-Day Spiritual Growth Journey-part-1


apps/backend/src/modules/plan-pdf-export/plan-pdf-export.service.ts
```diff
@@ -110,14 +88,98 @@ class PlanPdfExportService {
...
+			format: dto.format,
+			preferCSSPageSize: true,
+			printBackground: true,
+		});
+
+		return Buffer.from(pdfBuffer);
+	}
+
+	private async launchBrowser(): Promise<puppeteer.Browser> {
+		return await puppeteer.launch();
```
WilhelmWesser, line 131:
```suggestion
		return await puppeteer.launch({
		    headless: "new",
		    args: [
		        "--no-sandbox",
		        "--disable-setuid-sandbox"
		    ]
		});
```
Let's try it in this way to make it work on deploy. Please, let me know if this fix does not work locally

WilhelmWesser (changes requested):

apps/backend/src/modules/plan-pdf-export/plan-pdf-export.service.ts
muxkuro: It works, but the problem is that there is no "new" option. Only:
headless: true which launches the browser in a new headless mode
headless: "shell" which launches old headless mode (shell)

Which one of those will be better for you here?

WilhelmWesser: let's go with true

daniil-tkalenko (approved):

